PR #58, opened 2022-02-06
Clean up some CARES documentation

ajhyndman:
Also:
- add new ARPA treasury specification to repository
- replace "CARES" text with "ARPA" in code


igor47:
this change LGTM :+1: 

i guess the origin of this repo is a copy of the usdigitalresponse/cares-reporter repo? does the original repo still require maintenance, and who is doing that maintenance? i know there's some conversation about creating a single reporting tool that handles multiple reporting requirements, and i wonder how creating a new, separate repo plays with that plan. maybe that's for later, and the priority right now is getting arpa reporting working? sounds fine to me, but just wanted to note that if we continue maintaining the previous repo, then the maintenance burden will be doubled, e.g. for security patches that need to be made to both repos.

ajhyndman:
@igor47 yep, that's right, this repo was forked from cares-reporter.  I'm not very familiar with the cares project or its maintenance, so I can't comment on that.

I can note that this repository has been proposed as a home for the future, single consolidated tool.  https://github.com/usdigitalresponse/easy-config-reporter

If successful, I imagine that project could replace both arpa-reporter and cares-reporter.

PR #59, opened 2022-02-06
Add hot reloading support for express routes

ajhyndman:
I noticed that changes I made to API routes in the reporter were not detected without stopping and restarting the vue cli service.

Since I expect that a lot of the work we'll be doing in the next few weeks involves validation, I thought it was worth spending a little time to see if I could get a hot reloading workflow set up.

I introduced two new dev dependencies, but I was able to get it working without any major rewriting!  Let me know what you think.


igor47 (approved):
i think i'm following what's going on here, and it all makes sense. tried to double-check if any readme or test changes are necessary, and didn't seem like that's the case -- the readme says to run `yarn serve` and that still seems to be the case.

PR #74, opened 2022-02-10
Resolve lint errors!

ajhyndman:
The last commit in this pull request is the most interesting to review.

I attempted to resolve all of the remaining lint warnings, but decided it made more sense to disable ESLint in a few places, rather than reinvent the application architecture.

### DIFF
--- a/src/server/db/audit-report.js
+++ b/src/server/db/audit-report.js
@@ -8,10 +8,10 @@
 */
 const knex = require('./connection')
 
-let log = () => {}
-if (process.env.VERBOSE) {
-  log = console.log
-}
+// let log = () => {}
+// if (process.env.VERBOSE) {
+//   log = console.log
+// }
 
 module.exports = {
   getAggregateAwardData,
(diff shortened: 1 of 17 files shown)


src/server/db/audit-report.js
```diff
@@ -8,10 +8,10 @@
 */
 const knex = require('./connection')
 
-let log = () => {}
-if (process.env.VERBOSE) {
-  log = console.log
-}
+// let log = () => {}
```
igor47, line 11: what's the goal with commenting these lines out vs. just removing them?

igor47 (approved):
changes me sense to me, confirmed that first commit is the result of autofixes applied by just running `yarn lint` and the changes in the second commit also make sense, though i left one question.

opened #81 to maybe prevent this state from re-occuring

src/server/db/audit-report.js
ajhyndman: I don't have a strong opinion here, happy to just delete it.  I don't really know what this log helper is for.

I guess I was just anticipating that it may still be useful for debugging in the future.  If I delete it, I was just worried we would lose the context that this env variable is available to control logging levels.

Now that we've had this conversation, maybe that's enough that we'll both remember?  🤷

PR #84, opened 2022-02-18
Clean up commented-out logger

ajhyndman:
@igor47 I split off that last change you commented on from #74 .

Here I'm also adding the VERBOSE env variable to the .env.example file.  I feel like that could also help future developers discover this configuration?


igor47 (approved):
lgtm. i do see us using `log()` sometimes, and just straight-up `console.log` in other files, we could probably do a consistency pass

Issue #81, opened 2022-02-17
run lint as part of CI


igor47:
we got to a bunch of lint errors in the repo, fixed in #74 . this is probably because we're not automatically running lint on PRs/blocking PRs from being merged due to lint errors. could be something to look into

ajhyndman:
Resolved in #85!

PR #85, opened 2022-02-18
Add CI Workflow

Mr0grog:
This adds a continuous integration workflow that runs linters and tests on pull requests (and on pushes to `main` and `staging`) and reports the results as status checks. It should help make obvious that a PR is not safe before clicking the merge so we don’t have failed deploys or need things like #74 to quickly fix a bad merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,112 @@
+on:
+  pull_request: {}
+  push:
+    branches:
+      - main
+      - staging
+    paths-ignore:
+      - "doc/**"
+
+name: Continuous Integration
+
+jobs:
+  lint:
+    name: Run linters
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16
+          cache: 'yarn'
+
+      - name: Install dependencies
+        run: 'yarn install'
+
+      - name: Lint Source Code
+        run: |
+          yarn lint --no-fix
+
+  lint-workflows:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: cschleiden/actions-linter@v1
+        with:
+          workflows: '[".github/workflows/*.yaml", ".github/workflows/*.yml"]'
+
+  test-unit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16
+          cache: 'yarn'
+
+      - name: Install dependencies
+        run: 'yarn install'
+
+      - name: Run unit tests
+        run: |
+          echo "::warning file=package.json,title=Unit tests are broken::The 'yarn test:unit' command is currently broken"
+          # yarn test:unit
+
+  test-server:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: "postgres:13"
+        env:
+          POSTGRES_DB: arpa-reporter-test
+          POSTGRES_USER: test
+          POSTGRES_PASSWORD: test-password
+        ports:
+          - 5432:5432
+        # Set health checks to wait until postgres has started
+        options: >-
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,120 @@
...
+      postgres:
+        image: "postgres:13"
+        env:
+          # FIXME: the test scripts currently hardcode a database name, user,
+          # and password. The config here must match.
+          # (See tests/*/mocha_wrapper.sh)
+          POSTGRES_DB: 'server_test'
+          POSTGRES_USER: 'postgres'
+          POSTGRES_PASSWORD: 'password'
+          POSTGRES_HOST_AUTH_METHOD: 'trust'
```
Mr0grog, line 71: This issue is worth noting: while the app will take its database info from the environment variables, the *tests* hard-code a user and database name, and never use a password at all in the first place.

This configuration matches the hardcoded database name and user for the server tests (but not the `server-period-1` and `server-period-2` tests, which hardcode different names, and which I’m not sure are really meant to be used…?). It also sets `POSTGRES_HOST_AUTH_METHOD` to `trust` so that a password is not needed, since the tests hardcode not using a password. All of this is less than ideal, but at least it mostly works.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,120 @@
...
+
+      - name: Install dependencies
+        run: 'yarn install'
+
+      - name: Run server tests
+        run: |
+          # The .env file needs to be present; the example file is good enough.
+          cp .env.example .env
+          yarn knex migrate:latest
+          yarn test:server
```
Mr0grog, line 98: Using the `.env.example` file this way doesn’t feel especially great, but it works. It might be better to explicitly set all the environment variables in the `env:` YAML block just below this, but the test scripts are also coded to expect a `.env` file to always be present.

tests/server/mocha_wrapper.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-export POSTGRES_URL="postgres://localhost/server_test"
+export POSTGRES_URL="postgres://postgres@localhost/server_test"
```
Mr0grog, line 3: I had to add an explicit username here, otherwise the user is arbitrarily determined based on the linux user you are logged into the current linux machine as, which is something that might change from CI run to CI run. This happens to work as well for me locally (I use Postgres.app, and it always includes a `postgres` user in the database), but **it could break for other people.**

If this is a concern, we need to do some work so that this script loads some database configuration from the environment rather than hardcoding it all here.

Mr0grog (commented):
This *is* working now, although there are some things that are less than stellar here and which may or may not warrant more work before merging. I tried to note these concerns inline, but I’ll leave it up to reviewers to decide what should be done (or not done).

Mr0grog:
Once this lands, you may want to comment `@dependabot rebase` on all the pending Dependabot PRs. That way they’ll pick up on this new CI workflow and run all the tests.

tests/server/mocha_wrapper.sh
ajhyndman: This seems like a step forward to me.  I am happy to merge as is.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,120 @@
...
+      - main
+      - staging
+    paths-ignore:
+      - "doc/**"
+
+name: Continuous Integration
+
+jobs:
+  lint:
+    name: Run linters
```
ajhyndman, line 14: Nitpick: It looks like only one of these jobs has a name.  We could probably drop this name for consistency?

ajhyndman (approved):
This looks great to me!  Thanks very much for taking the time to set it up, @Mr0grog.

ajhyndman:
fixes #81

Mr0grog:
Looks like the server tests may be a bit flaky. :(

Otherwise this should be good to go! I’ll go ahead and merge later this afternoon if I don’t hear any concerns from you about the flakiness.

ajhyndman:
I'm happy to merge!  We may end up rewriting the "server tests" entirely, anyway.

PR #86, opened 2022-02-18
Limit Dependabot pull requests to lockfile changes

ajhyndman:
I merged all open dependabot changes a week ago, but immediately there was a new backlog to resolve.  In addition, several of the new dependencies introduced breaking changes that needed to be resolved by hand.

None of these version updates address meaningful security concerns, so I propose tweaking the dependabot configuration such that it respects the semver declarations in this project's `package.json`.

e.g. After this config change, if we declare a dependency on `"vue": "^2.6.14"`, dependabot will alert us to future minor versions or patches (e.g. `2.7.0` or `2.6.15`), but will not prompt us to upgrade across a major version (`3.2.31`).

My theory is that this will make it much easier for us to merge security patches and bug fixes in a timely fashion.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,9 @@ updates:
     schedule:
       interval: "daily"
     target-branch: staging
+    # Don't automatically suggest version updates that conflict with the
+    # package.json specification. This greatly reduces the likeihood that a
+    # dependabot update will break anything.
+    # When the package.json needs to be updated for security purposes, Github
+    # will alert us via a "security advisory" instead.
+    versioning-strategy: lockfile-only


igor47 (approved):
looked at https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#versioning-strategy to see what this config option does.

i was confused about the PR comment that `^1.2.3` would update to `^1.3.0` (i would expect only updates to `1.2.x`) but this appears true in  npm, verified by checking https://semver.npmjs.com/ 

agree that limiting the noise from dependabot would be quite nice

ajhyndman:
Yeah, "caret ranges" (^) basically enforce that the most-significant non-zero value in the version must match.

I think the idea is that upgrading from 0.1.0 to 0.2.0 is probably not safe, but 1.1.0 to 1.2.0 should be safe.  i.e. Minor version releases are meant to only add to an API in non-breaking ways, but that's not necessarily true for packages that have not had a 1.0.0 release.

Mr0grog:
FWIW, my concern on this is that the NPM package ecosystem tends to a) move pretty fast and b) not backport fixes and security updates, so if you get behind on major versions, you may never receive a security PR from Dependabot because there’s no compatible way for it to update to a secure version of whatever dependency has a problem. It’s dumb, but I worry about this/run into this problem way less in other package management ecosystems (e.g. apt, PyPI, Ruby Gems, etc.).

In general, I think better fixes for this issue are things like:
- Having CI checks on your PRs that do builds, run tests, run linters, etc. (which we have now done).
- Reducing the frequency of dependency checks so I deal with updates less often, e.g. weekly or monthly. (In projects I maintain long-term, I *never* set it to daily.)

That said, I totally hear you if even those approaches create too much noise and overhead for you! If never getting PRs for major version updates really would help the project move dramatically faster and more freely, maybe it’s worthwhile. :)

ajhyndman:
I'm going to land this.  Every time I merge a dependabot PR, three more are opened.  🙈

PR #95, opened 2022-02-22
Disable flaky test suite

ajhyndman:
@Mr0grog I isolated the flakiness to this test suite — which appears to depend on the order in which unit tests are executed.  I think disabling this test until we have a chance to rewrite it makes sense.


Mr0grog (approved):
Ah, yeah, I had noticed the same, but somehow foolishly assumed something had been done to get a predictable order. Obviously not! 🙃

PR #104, opened 2022-02-27
Fix hot reload firefox

ajhyndman:
I think we can get hot reloading and Firefox support together after all!

It seems like this line in vue-cli-service is to blame:

https://github.com/vuejs/vue-cli/blob/master/packages/%40vue/cli-service/lib/util/prepareProxy.js#L53

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -1,11 +1,10 @@
 module.exports = {
   devServer: {
-    proxy: 'http://localhost:3000',
-    client: {
-      // Set the websocket path to sockjs-node so that vue-cli-service knows
-      // not to proxy the request.
-      // https://github.com/vuejs/vue-cli/blob/master/packages/%40vue/cli-service/lib/util/prepareProxy.js#L53
-      webSocketURL: { pathname: '/sockjs-node' }
+    proxy: {
+      // proxy requests whose paths match the pattern below to express
+      '/api': {
+        target: 'http://localhost:3000'
+      }
     }
   }
 }
--- a/vue.config.js
+++ b/vue.config.js
@@ -1,6 +1,11 @@
-
 module.exports = {
   devServer: {
-    proxy: 'http://localhost:3000'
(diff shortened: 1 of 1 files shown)


ajhyndman:
@igor47 It'd be great if you can verify whether this fixes the issue for you in Firefox!

igor47:
the app still doesn't work for me locally on this branch. i can request a login link, but i can't log in. it *does* fix the web socket error, but that seems to be unrelated. 

i ask for the login link at http://127.0.0.1:8080/login
i get a link that sends me to, e.g., http://127.0.0.1:8080/api/sessions?passcode=c750fb16-9565-4969-aa3d-16d205d34467
if i access that URL, i see that the token is not marked `used` in the DB; also, my printf debugging shows that no url params are received: `[dev:*express] params are undefined`

if i change the URL to go to port 3000 instead of 8080, then the params *are* used, but i cannot use the site with express directly -- `/` is a 404, and of course the cookie i get from logging in then applies to the wrong server (at 3000 instead of 8080). i think there is some problem with how the vue server proxies to express, but i'm not sure where.

ajhyndman:
Ooh, now we're getting closer.  I can also repro if I attempt to load the app from `http://127.0.0.1:8080`.  Does the `localhost` alias work on your machine?  I have been using `http://localhost:8080` as my browser URL, which appears to work.

It seems like the app shouldn't care about the difference there, but maybe that will get you running?

ajhyndman:
@igor47 I pushed another change with a different workaround.

Instead of changing the socket path to prevent it being accidentally proxied, we instead restrict proxying to paths that start with `/api`.  I can't fully explain it, but this also appears to fix the issue we observe when loading thee site at http://127.0.0.1.  🤞

igor47 (approved):
this works for me now! i'm content to not dig into what was happening here, esp. since this is a dev-only change and everything is totally different in production

PR #151, opened 2022-03-31
Migration Scripts

igor47:
the goal of this PR is to make it easy to stand up new instances of render for ohio/ri/other customers we might want. it's primary designed to (1) make sure we have environment variables set and (2) copy existing agencies and users from the cares reporter

while we're at it, we clean up some of the seed files so they're more relevant, and so that the seeds can be re-run without destroying all data.

the flow goes like this. follow along as I stand up a new instance based on an old instance:
* create a new postgres service in render. use postgres version 13
* create a new web service using `usdigitalresponse/arpa-reporter` (use the branch `updated_seeds` for now) -- the build command is `yarn install && yarn build` and the start command is `yarn run prod`. in the `advancated` tab, create a disk, name `Uploads`, mount path `/var/data/uploads`. set the health check path to `/api/health`. the plan must be at least `Starter` to get SSH access
* wait for the service to deploy
* on your local laptop, add your render API key to your environment and then run `yarn initRender` -- select your new service as the service you're configuring, and select an existing instance as a source, if applicable. follow the prompts to make sure you configure all environment variables. for `POSTGRES_URL`, you'll need to go into the postgres service you created and copy the line under `Internal Connection String`. `POSTGRES_SOURCE_URL` should be the DB instance of your source, where additional users and agencies will be copied from
* after you get the output `Successfully updated environment`, re-deploy the web service to get it to pick up the new environment variables
* now configure the db. in the web service, click on `Shell`, and run `yarn knex migrate:latest` and `yarn knex seed:run`
* you should now be able to log into your new instance. you should see that all the agencies as well as all the users are copied over.

### DIFF
--- a/seeds/05_app_settings.js
+++ b/seeds/05_app_settings.js
@@ -1,8 +1,17 @@
 require('dotenv').config()
 
+const title = (process.env.INITIAL_APP_TITLE || 'ARPA Reporter')
+
 exports.seed = async function (knex) {
-  await knex('application_settings').del()
+  const [{ count }] = await knex('application_settings').count('title', { as: 'count' })
+  if (count !== '0') {
+    console.log('db already has application settings...')
+    return
+  }
+
+  const { firstPeriod } = await knex('reporting_periods').first('id AS firstPeriod')
+
   await knex('application_settings').insert([
-    { title: 'Rhode Island', current_reporting_period_id: 2 }
+    { title, current_reporting_period_id: firstPeriod }
   ])
 }
--- a/seeds/05_app_settings.js
+++ b/seeds/05_app_settings.js
@@ -1,8 +1,8 @@
-require("dotenv").config();
+require('dotenv').config()
 
-exports.seed = async function(knex) {
-  await knex("application_settings").del();
-  await knex("application_settings").insert([
-    { title: "Rhode Island", current_reporting_period_id: 2 }
-  ]);
-};
+exports.seed = async function (knex) {
+  await knex('application_settings').del()
+  await knex('application_settings').insert([
+    { title: 'Rhode Island', current_reporting_period_id: 2 }
+  ])
+}
--- a/seeds/06_migrate_from_source.js
+++ b/seeds/06_migrate_from_source.js
@@ -37,10 +37,12 @@ exports.seed = async function (knex) {
   })
 
   // now users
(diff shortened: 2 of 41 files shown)


igor47:
@jcomeau1 i kind of guessed on the reporting periods. i see that some organizations report quarterly and some report annually, i'm not sure which reporting periods we should have in the DB. see this file: https://github.com/usdigitalresponse/arpa-reporter/pull/151/files#diff-413a00565138162a00e4c349d138f56d0973866fd9dc808098c0b4e247498542

igor47:
probably, after we deploy the new instances, we should upload the correct template for the reporting period. this is done by going to "Reporting Periods" in the UI and hitting "Upload template". i don't think the code currently lets you download a template without it.

seeds/05_app_settings.js
```diff
@@ -1,8 +1,17 @@
...
+  if (count !== '0') {
+    console.log('db already has application settings...')
+    return
+  }
+
+  const { firstPeriod } = await knex('reporting_periods').first('id AS firstPeriod')
+
   await knex('application_settings').insert([
-    { title: 'Rhode Island', current_reporting_period_id: 2 }
+    { title, current_reporting_period_id: firstPeriod }
```
ajhyndman, line 15: Oh, this is definitely nicer.  👍

src/server/configure.js
```diff
@@ -33,6 +33,7 @@ module.exports = app => {
   app.use('/api/uploads', require('./routes/uploads'))
   app.use('/api/users', require('./routes/users'))
   app.use('/api/validations', require('./routes/validations'))
+  app.use('/api/health', require('./routes/health'))
```
ajhyndman, line 36: 👍

seeds/06_migrate_from_source.js
```diff
@@ -0,0 +1,57 @@
+require('dotenv').config()
```
ajhyndman, line 1: Is the idea of this script that you're cloning agencies and users from CARES/CRF reporting?

I assume there will be some overlap between agencies involved in CARES and in ARPA, but I wonder if we should run that by the states before cloning everything wholesale.

ajhyndman (approved):
Thanks for putting this together!  It looks super useful.

I left a couple comments, but I don't think there's anything blocking.

PR #159, opened 2022-04-06
Remove CARES-specific integration tests

ajhyndman:
This removes a set of integration style tests that assert behaviours about the input and export spreadsheet formats of CARES reporting.


igor47 (approved):
lgtm! lets re-write these as more unit tests. maybe we can do `POST`s to the API to do upload validation, and our unit tests for the output generation can directly manipulate the DB and ask for a CSV generation and validate the resulting CSV

PR #173, opened 2022-04-11
Parse uploads

ajhyndman:
Parse an uploaded spreadsheet and save its data into the documents table.

This is blocked on #158 and #159. Note that this diff does _not_ yet attempt to support reading the data back.


src/server/lib/spreadsheet.js
```diff
@@ -79,12 +80,20 @@ function parseSpreadsheet (workbook, templateSheets) {
...
+  // remove metadata rows & normalize column names
+  const parsedWorkbook = {}
+  Object.entries(jsonWorkbook).forEach(([sheetName, sheet]) => {
+    const normalizedSheet = removeMetadata(sheetName, jsonWorkbook[sheetName])
+    const [headerRow, ...rest] = normalizedSheet
+    const normalizedHeader = _.map(headerRow, colName =>
+      typeof colName === 'string' ? colName.toLowerCase().trim() : colName
+    )
+    parsedWorkbook[sheetName] = [normalizedHeader, ...rest]
+  })
```
igor47, line 96: this is a great candidate for a unit test; it could be broken out into a separate function for ease of testing

src/server/services/process-upload.js
```diff
@@ -102,7 +101,8 @@ const processUpload = async ({
           docs.push(doc)
         }
       })
-      subrecipients.update(subs)
+      // FIXME: Save subrecipients data to subrecipients table?
+      // subrecipients.update(subs)
```
igor47, line 105: not sure if this is the place to comment this out. are you having problems saving the sub-recipients? are they in a different schema?

igor47 (approved):
looks good, i think i understand the sub-recipients comment-out, we can address that in a follow-up PR

PR #180, opened 2022-04-13
Commit export templates

ajhyndman:
Save the full set of export templates (a.k.a. bulk upload templates) to the repository.

NOTE: We could convert all of these files as .csv before saving to the repository.  However, I believe that the XLSX library reads both formats into memory in the same generic interchange format, so I don't think there's actually much advantage?

This pull request also makes the additional change of extracting the `latestReport` utility function to its own file.  Now `lib/treasury.js` is not imported by anything other than `routes/exports.js`.


src/server/services/get-template.js
```diff
@@ -11,6 +11,10 @@ const xlsx = require('xlsx')
 const _ = require('lodash')
 const { sheetToJson } = require('../lib/spreadsheet')
 
+// Treasury upload spreadsheet template
+// TODO: Support all treasury export templates
+const PROJECT_BASELINE_TEMPLATE = 'treasury/projectBaselineBulkUpload.xlsx'
```
ajhyndman, line 16: Based on our chat earlier, I think we want to just configure the export templates in code, rather than via env variable.  What do you think?

igor47: sgtm; i wonder if we want to declare a mapping from type of output to it's filename, vs having a variable name for each of the 16 output templates? then `loadTreasuryTemplate` can take a `type arg for the output template to load

src/server/lib/latest-report.js
```diff
@@ -0,0 +1,46 @@
+/* eslint camelcase: 0 */
```
igor47, line 1: can probably kill this line if you rename `period_id` to `periodId`

igor47 (approved):
lgtm! left some minor comments

src/server/lib/latest-report.js
ajhyndman: Yeah, I just decided to leave it alone for consistency with other usages through the codebase.  🤷‍♂️

src/server/services/get-template.js
ajhyndman: That sounds pretty reasonable to me.  This is definitely just a placeholder.  👍

PR #183, opened 2022-04-14
Igor/generate arpa report zip

igor47:
This PR contains the scaffolding code for treasury report generation. We re-enable the `Download Treasury Report` button, and have that button generate a zip file containing the template CSV files we get from the treasury. The output templates are generated in functions -- one function per output file type -- that serve as the scaffolding for implementing the business logic. As a next step, we should actually populate the CSV files with data from the documents uploaded by agency users.

We remove all the old treasury-related code from CARES, including tests. We should implement new testing for our new output file generation.

We implement a new centralized configuration location, in `src/server/environment.js`, which aims to include all environment-specific configuration options.

Testing was broken in my environment using docker, because it assumed a specific username/password. We update the mocha wrapper scripts here to hopefully make testing work on any machine, by deriving a test db from the dev db/dev credentials.


tests/server/reset-db.sh
```diff
@@ -11,13 +14,27 @@ password=${userpass#*:}
...
+fi
+
 echo Using database $dbname
 
 mkdir -p $UPLOAD_DIRECTORY
 rm -rf $UPLOAD_DIRECTORY/*
 
-PGPASSWORD=$password psql -h $host -U $username -w ${DEVDBNAME} -c "DROP DATABASE IF EXISTS $dbname"
-PGPASSWORD=$password psql -h localhost -U postgres -w ${DEVDBNAME} -c "CREATE DATABASE $dbname"
+set +x
```
ajhyndman, line 29: Should be `set -x`

ajhyndman (commented):
Cool, let's land it!

ajhyndman (approved):

PR #186, opened 2022-04-15
Move vue/cli and eslint deps to "devDependencies"

ajhyndman:
Declares vue CLI and eslint dependencies as devDependenices.

This is useful because auditing tools treat these categories differently.

before:

image

after:

image


igor47 (approved):
lgtm!

PR #188, opened 2022-04-15
Update empty template to latest version from Joe

igor47:
he says this is the final final one


src/server/environment.js
```diff
@@ -14,7 +14,7 @@ const ARPA_REPORTS_DIR = join(DATA_DIR, 'arpa_reports')
 const SRC_DIR = resolve(join(__dirname, '..'))
 const SERVER_DATA_DIR = join(SRC_DIR, 'server', 'data')
 
-const EMPTY_TEMPLATE_NAME = 'empty-template.xlsm'
+const EMPTY_TEMPLATE_NAME = 'ARPA SFRF Reporting Workbook v20220414.xlsm'
```
ajhyndman, line 17: Are you wanting to keep a history of input templates committed to the repository?

I'd rather either delete empty-template.xlsx or rename it to match this naming convention.

ajhyndman (changes requested):

src/server/environment.js
igor47: i think it might be helpful for users to know which version of the template they're downloading. i've deleted the old `empty-template.xlsm`

ajhyndman (approved):

PR #190, opened 2022-04-15
Use XLSX package to stringify CSV file

ajhyndman:
This just lets us keep one fewer dependency around.


igor47 (approved):
oh cool thanks, couldn't figure out how to do that

PR #191, opened 2022-04-15
Populate project baseline template for export

ajhyndman:
To manually run through the tool:

- `yarn serve`, open browser and log in
- In the **Agencies** tab, create a new agency with ID `OMB01`
- In the **Projects** tab, create a new project with code `768622`
- From **Dashboard**, click **Upload Agency Spreadsheet** and select `tests/server/fixtures/arpa-examples/OBM01-768622-09302020-v20220320.xlsm`
- Select **Dashboard**, click **Download Treasury Report**
- Extract zip file, data should be populated in `projectBaselineBulkUpload.csv`


src/server/services/generate-arpa-report.js
```diff
@@ -2,11 +2,11 @@
 const path = require('path')
 const { mkdir, rmdir, writeFile, readdir, readFile } = require('fs/promises')
 const moment = require('moment')
-const { stringify } = require('csv-stringify/sync')
 const zipper = require('zip-local')
 const XLSX = require('xlsx')
 
 const { applicationSettings } = require('../db/settings')
+const { documentsWithProjectCode } = require('../db')
```
igor47, line 9: hmmm is there a more specific place to require this from? i'm not sure how i feel about re-exporting so much stuff in the db index

src/server/services/generate-arpa-report.js
```diff
@@ -37,7 +43,7 @@ async function loadTemplate (templateName) {
   }
 
   const worksheet = workbook.Sheets[workbook.SheetNames[0]]
-  return XLSX.utils.sheet_to_json(worksheet, { header: 1 })
+  return XLSX.utils.sheet_to_json(worksheet, { header: 1, blankrows: false })
```
igor47, line 46: i remember you saying that this sometimes messes with things, like if we expect a specific cell at a specific index and then the indexing is thrown off after blank rows are removed. should this be the default?

src/server/services/generate-arpa-report.js
```diff
@@ -81,7 +87,64 @@ async function generateProject519521 (periodId) {
...
+  documents.forEach(document => {
+    console.log(document)
+
+    switch (document.type) {
+      case 'certification':
+        break
+      case 'cover':
+        rowObject.Project_Expenditure_Category__c = document.content['expenditure category group']
+        rowObject.Sub_Category__c = document.content['detailed expenditure category']
+    }
```
igor47, line 111: instead of these switch statements, can we `filter`? or possibly even ask the `documentsWithProjectCode` for only the types of documents we care about?

src/server/services/generate-arpa-report.js
```diff
@@ -81,7 +87,64 @@ async function generateProject519521 (periodId) {
...
+  const rowObject = {}
+
+  documents.forEach(document => {
+    console.log(document)
+
+    switch (document.type) {
+      case 'certification':
+        break
+      case 'cover':
+        rowObject.Project_Expenditure_Category__c = document.content['expenditure category group']
```
igor47, line 109: are subsequent documents meant to clobber this property on the rowObject ?

igor47 (commented):
just leaving some comments on the draft. i started doing that before i realized it was a draft :stuck_out_tongue_closed_eyes:

src/server/services/generate-arpa-report.js
ajhyndman: It seems that the xlsx export template files have 999 rows, mostly empty.  If we don't strip empty rows, then any new data we append will start at row 1000 — I think that's likely to be confusing to either humans or the treasury portal.

The issue I saw in the past, was with blank row elimination during read of input files.  We could try the inverse setting on the file parser for input sheets; that's not a bad idea.  👍

ajhyndman:
@igor47 I should clarify that this PR is just a draft!  I'm not happy with most of the structure either.  I just wanted to share WIP with one workflow "working".

src/server/lib/spreadsheet.js
```diff
@@ -172,6 +175,21 @@ async function spreadsheetToDocuments (
...
+          break
+
+        case 'ec 1 - health':
+        case 'ec 2 - economic impact':
+        case 'ec 3 - services':
+        case 'ec 4 - premium pay':
+        case 'ec 5 - infrastructure':
+        case 'ec 7 - admin':
+          jsonRow['detailed expenditure category'] = detailedExpenditureCategory
+          break
```
ajhyndman, line 191: This is definitely weird.  Let me know if you can think of a cleaner approach to associate detailed expenditure category with individual project "documents".

ajhyndman:
Superceded by #251

PR #198, opened 2022-04-21
Latest blank template from Joe

igor47:
follow-up to #188 with another update. leaving both versions in the repo now that the files are property versioned, could go either way on that.


ajhyndman (approved):
🙌

PR #200, opened 2022-04-21
parse-filename: Support project codes with hyphens

ajhyndman:
Our example data already includes the example of a project code including a hyphen `768622-CVR`.

This change switches our filename parsing from splitting file parts on hyphens to matching the entire filename with a regular expression.

### DIFF
--- a/src/server/services/parse-filename.js
+++ b/src/server/services/parse-filename.js
@@ -8,6 +8,15 @@ const { ValidationItem } = require('../lib/validation-log')
 const { agencies, agencyByCode, projectByCode } = require('../db')
 const { format } = require('date-fns')
 
+// To more clearly understand this regex, checkout this railroad diagram visualization:
+// https://regexper.com/#%2F%5E%28%5Ba-zA-Z%5D%2B%29-%28.%2B%29-%28%5B0-9%5D%7B8%7D%29-%28%5B%5E-%5D%2B%29%24%2F
+// Capture groups should be:
+// 1. Agency Code (alphanumeric)
+// 2. Project Code (arbitrary string)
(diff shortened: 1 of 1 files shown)


igor47 (approved):
this PR seems fine, however, is this the only way to get some of the data?

in some places in the code, we seem to assume that agency code for the upload has to do with agency code for the user, but that's not always set. i guess deeper down in the logic, we extract the agency code from the filename and set that as the document agency code?

for project code -- what is project code? this is not already somewhere IN the document?

for reporting period, we seem to assume at least in some places that the period is the currently open period, why do we need that in the filename?

i'm not sure the purpose of the version.

if the goal is to disambiguate the filenames on the server side filesystem, my plan currently is to just save whatever filename was given when uploading (this is available in the `uploads` table) and then persist the document in the server filesystem using it's unique upload id. this allows multiple uploads with the same client-side name to be successfully uploaded with unique server-side names.

if the goal is to get some information that is not available *IN* the spreadsheet, i propose adding fields to the upload form for this. adding an agency drop-down would be easy enough if necessary for output parsing. adding a drop down for reporting period id is likewise an already-requested feature (users want to upload documents for arbitrary reporting periods). not sure about project code.

ajhyndman:
Yeah, you raise some great questions!

Now that I stop and think again, I recall that Joe pointed out yesterday that we will at minimum need to figure out how to remove the project code from the filename.  This is because a single upload spreadsheet might include data for multiple projects.

I think we should review whether enforcing a naming convention is still useful with partners tomorrow.

ajhyndman:
I can't figure out how to untangle all of the logic around project ID tonight so I think I'm going to land this, but we do need to follow up!

PR #207, opened 2022-04-26
Igor/uploads overhaul

igor47:


src/server/services/process-upload.js
```diff
@@ -1,124 +1,86 @@
...
 
-  if (!valog.success()) {
-    log('valog.success() is false')
-    return { valog, upload: {} }
+  // get begin generating (partial) document rows
+  const documents = []
+  for (const sheetName of workbook.SheetNames) {
+    documents.push({
+      type: normalizeSheetName(sheetName),
+      content: xlsx.utils.sheet_to_json(workbook.Sheets[sheetName], { header: 1 })
```
igor47, line 32: @ajhyndman -- i'm not sure if this is correct/preferred, but here i'm just writing the entire spreadsheet into the document row. i know that previously, we tried to parse the document here, and save it with -- column headings? not sure what the previously format is. i'm wondering if this makes sense. i do think we can transform the sheet on read when producing treasury reports, and keeping it the rows/columns just as input would make validation a little easier, we can point the user to the exact cells that fail to validate by the row/column number

src/server/services/validate-upload.js
```diff
@@ -1,105 +1,82 @@
...
-    workbookXlsx = xlsx.read(data, { type: 'buffer' })
-  } catch (e) {
-    console.log('error', e)
-    valog.append(`Can't parse xlsx file ${filename}`)
-    return { valog, documents: {} }
-  }
+async function validateAgencyId ({ upload, documents }) {
+  // grab agency id from the cover sheet
+  const coverSheet = documents.find(doc => doc.type === 'cover').content
+  const agencyCode = coverSheet[1][0]
```
igor47, line 29: @ajhyndman example of how to pull data out of the spreadsheet now that it's just an array of arrays, the error below can point to the same location as where i'm looking

src/server/services/process-upload.js
ajhyndman: Oh, interesting!

I think at some point we're going to have to do the work we were doing here, regardless.  In the old system, the "documents" table was actually a collection of "rows".  Ultimately, we're going to need to parse the spreadsheets and extract rows so that we can aggregate across uploads at write time.

I think my first instinct is to keep doing that and keep this table.  Does that cause problems for your new upload approach?

In some places in the code, these are called "records" instead of "documents" — that might be a better name for the table, if we keep it.

src/server/services/validate-upload.js
ajhyndman: If we do go this route, I think I'd rather save the data in some format with richer metadata than array of arrays.  Trying to access or manipulate data by column and row position sounds pretty error-prone to me.

I believe sheetjs (XLSX library) has support for a "JSON" data type as well.  This saves rows as objects with named keys (derived from headers).  The API is weird, from reading the docs, I think you _omit_ the header option from sheet_to_json to get this behaviour?

Alternatively, we could probably just save the default sheetjs interchange format?

src/server/routes/uploads.js
```diff
@@ -84,4 +124,22 @@ router.get('/download/:id', requireUser, (req, res) => {
   })
 })
 
+router.get('/:id/validate', requireUser, async (req, res) => {
```
ajhyndman, line 127: I think I'd like to make this a POST request, for clarity.

PR #210, opened 2022-04-28
Project baseline on uploads overhaul

igor47:
a version of #191 based on the new uploads format in the `documents` table


ajhyndman:
I think we can close this PR now, correct?  Is there anything you want to salvage first, @igor47 ?

PR #212, opened 2022-04-29
Igor/kill documents

igor47:


src/server/services/persist-upload.js
```diff
@@ -83,4 +68,17 @@ async function persistUpload ({ filename, user, buffer }) {
   return upload
 }
 
-module.exports = { persistUpload, ValidationError, uploadFSName }
+async function bufferForUpload (upload) {
+  const data = await readFile(uploadFSName(upload))
+  return Buffer.from(data, 'binary')
+}
+
+async function documentsForUpload (upload) {
```
ajhyndman, line 76: I think we should try to rename any remaining references from "documents" to "records".

I guess it probably makes more sense to make that a follow-up task though.

src/server/lib/validation-error.js
```diff
@@ -0,0 +1,22 @@
+
+class ValidationError extends Error {
```
ajhyndman, line 2: Just curious: Is this the file you had to extract to break the circular dependency?

src/router/index.js
```diff
@@ -33,12 +31,6 @@ const routes = [
     component: Home,
     meta: { requiresLogin: true }
   },
-  {
-    path: '/documents/:type',
```
ajhyndman, line 37: I guess this route is the most compelling argument _against_ removing this table, right?  We won't be able to easily support these views anymore without the old "documents" table.

I think I still want to go ahead.  If it turns out there was a compelling use case we're no longer serving, we can probably think up other ways to tackle it as well.

Actual feedback:

We should also remove the nav links associated with this route!

https://github.com/usdigitalresponse/arpa-reporter/blob/bbb9d4ac7c986f6e38c65266e82a0187a8625feb/src/components/Navigation.vue#L51-L71

ajhyndman (changes requested):
I just have one significant change request:

We should also clean up the nav links for the /documents/:id route that you removed!

src/server/lib/validation-error.js
igor47: yup

src/server/services/persist-upload.js
igor47: well, it now actually is returning documents, though maybe it could be `filesForUpload`

src/router/index.js
igor47: done

ajhyndman (approved):
Thanks!  I'll unblock this, let's get it landed.

PR #220, opened 2022-04-30
Cache treasury templates in memory

ajhyndman:
This makes the download workflow _significantly_ faster!  It seems like reading and parsing the template files is the slowest part of this process.

image

### DIFF
--- a/src/server/services/get-template.js
+++ b/src/server/services/get-template.js
@@ -9,16 +9,11 @@ const path = require('path')
 const fs = require('fs')
 const xlsx = require('xlsx')
 const _ = require('lodash')
+const { SERVER_DATA_DIR } = require('../environment')
 const { sheetToJson } = require('../lib/spreadsheet')
 
-// Treasury upload spreadsheet template
-// TODO: Support all treasury export templates
-const PROJECT_BASELINE_TEMPLATE = 'treasury/projectBaselineBulkUpload.xlsx'
-
-const treasury = {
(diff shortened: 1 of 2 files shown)


src/server/services/get-template.js
```diff
@@ -28,7 +23,7 @@ const validation = {
 
 module.exports = {
   getDropdownValues,
-  getTreasuryTemplateSheets,
+  getTemplate,
```
igor47, line 26: nit: i would have named `getTemplate` as `loadTemplateCached` and then exported `loadTemplate: loadTemplateCached` to make it clearer what's happening, this would have prevented changes downstream as well

igor47 (approved):
lgtm, left one nit.

i am a little concerned that this is so slow. now that we killed the `documents` table, we'll be loading and parsing many, many uploads during report generation. we'll see how it goes! i wonder if retrieving a binary file is faster from the DB than the filesystem? or is it deserialization that's the bottleneck? (xlsx seems pretty slow).

src/server/services/get-template.js
ajhyndman: I generally like verbose names too.  I just went with getTemplate, because I noticed that all of the functions in this file use an existing pattern:

- `loadX` actually calls the filesystem
- `getX` wraps `loadX` with caching logic

I don't really want to rename everything in this change, but I'm happy to do so afterward!

PR #221, opened 2022-04-30
Create exports zip file in-memory

ajhyndman:
[OPTIONAL]

This change uses adm-zip (a js implementation of the zip algorithm) to generate our exported zip file without having to interact with the disk.

I suspected this would meaningfully improve our export times, but I was not able to measure a meaningful performance change.  It's possible that this will change with larger quantities of real data.

The remaining advantages of this approach are:
- we never have to worry about accumulating intermediate export files on disk, and we don't have to think about cleaning those files up
- it fixes two [security](https://github.com/usdigitalresponse/arpa-reporter/security/dependabot/26) [vulnerabilities](https://github.com/usdigitalresponse/arpa-reporter/security/dependabot/27) that dependabot is complaining about.

🤷  I don't think this is as compelling as #220 yet, but I figured I might as well push the change and talk it through.

### DIFF
--- a/src/server/services/generate-arpa-report.js
+++ b/src/server/services/generate-arpa-report.js
@@ -102,9 +102,7 @@ async function generateSubRecipient (periodId) {
   return loadTemplate('subawardBulkUpload')
 }
 
-async function generateReport(periodId) {
-  console.log('generateReport called', performance.now())
-
+async function generateReport (periodId) {
   // generate every csv file for the report
   const csvObjects = [
     { name: 'project18_229233BulkUploads', func: generateProject18 },
--- a/src/server/services/generate-arpa-report.js
+++ b/src/server/services/generate-arpa-report.js
@@ -132,8 +132,6 @@ async function generateReport (periodId) {
     { name: 'subRecipientBulkUpload', func: generateSubRecipient }
   ]
 
-  console.log('csvFiles generated', performance.now())
-
   const zip = new AdmZip()
 
   // compute the CSV data for each file, and write it into the zip container
--- a/src/server/services/generate-arpa-report.js
+++ b/src/server/services/generate-arpa-report.js
@@ -1,8 +1,7 @@
-
 const path = require('path')
-const { mkdir, rmdir, writeFile, readdir, readFile } = require('fs/promises')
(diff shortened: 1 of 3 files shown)


src/server/services/generate-arpa-report.js
```diff
@@ -125,42 +116,53 @@ async function generateReport (periodId) {
     { name: 'project51518BulkUpload', func: generateProject51518 },
     { name: 'project519521BulkUpload', func: generateProject519521 },
     { name: 'projectBaselineBulkUpload', func: generateProjectBaseline },
-    { name: 'expendituresGT50000BulkUpload', func: generateExpendituresGT50000 },
-    { name: 'expendituresLT50000BulkUpload', func: generateExpendituresLT50000 },
-    { name: 'paymentsIndividualsLT50000BulkUpload', func: generatePaymentsIndividualsLT50000 },
+    {
```
igor47, line 119: do you have a max line length setting in your editor? we should agree on max line length, i don't think one is enforced right now.

src/server/services/generate-arpa-report.js
```diff
@@ -125,42 +116,53 @@ async function generateReport (periodId) {
...
-    .catch((err) => {
-      rmdir(dirName, { recursive: true })
-      throw err
-    })
-
-  // now we generate a zip file
-  const zipfileName = dirName + '.zip'
-  const zipfile = zipper.sync.zip(dirName)
-  zipfile.compress().save(zipfileName)
+  console.log('csvFiles generated', performance.now())
```
igor47, line 135: did you mean to leave this in?

igor47 (approved):
lgtm, left some nits. nice to not deal with the filesystem here.

src/server/services/generate-arpa-report.js
ajhyndman: Yeah, good question.

I don't have a strong personal preference here.  What do you think about adopting prettier checks in CI and removing conflicting eslint rules to resolve this?  Is prettier auto-formatting easy to work with your IDE?

src/server/services/generate-arpa-report.js
ajhyndman: Whoops, thanks for the catch!

PR #225, opened 2022-05-04
Further Cleanup to remove DataTable

igor47:
The goal here was to remove the DataTable component, and it's various confusingly-named sub-components. It's remaining uses in the codebase were as follows:

* in the `Agencies` view, we switched to using normal plain HTML table
* the `Projects` view was completely removed, along with it's backend functionality -- routes, db queries, and the `projects` DB table
* the `UploadHistory` UI was completely re-written to use normal tables, and also to manage it's own state from the backend instead of depending on being passed props containing uploads from a parent component

We still have a `FastTable`, used in the `Subrecipients` UI, but I'll leave that cleanup for a subrecipients-focused follow-up PR.


tests/unit/components/UploadHistory.spec.js
```diff
@@ -17,15 +17,15 @@ describe('UploadHistory.vue', () => {
...
-      stubs: ['router-link'],
-      propsData: {
-        uploads: [{ filename: 'one.xlsx' }, { filename: 'two.xlsx' }]
-      }
+      stubs: ['router-link']
     })
-    expect(wrapper.text()).to.include('Upload')
+    expect(wrapper.text()).to.include('No  uploads')
+
+    /* TODO: figure out how to intercept fetch api requests
```
ajhyndman, line 24: If we're consistent, I think we should be able to mock this file in tests:

https://github.com/usdigitalresponse/arpa-reporter/blob/staging/src/store/index.js

I'm not as familiar with mocha+chai, but it looks like this plugin might be the way to go?

https://www.chaijs.com/plugins/chai-spies/

ajhyndman (commented):
Looks pretty good!  I want to register two comments for discussion.

1. I'm all for removing DataTable (we don't need a hand-rolled solution to render a table), but I want to watch out that we don't remove 3 JS files and add a megabyte of 3rd party code.
2. I'd really rather keep uploads in the store here.  There's a little extra indirection, but I think there's value in being able to cache the request and better isolate our UI components for unit testing.  Was there something about this change that made that hard to preserve?

ajhyndman (approved):
We can review follow ups in #233 if you prefer!

PR #233, opened 2022-05-10
EC Code tracking

igor47:
This begins tracking uploads by EC code. The UI is changed so that EC codes are displayed on the various upload tables. The backend will now return uploads based on [reporting period, ec code, agency].


igor47:
this PR is opened against #225 and alas includes some changes requested in that PR. i recommend merging this into 225 and then merging both into staging together.

src/main.js
```diff
@@ -5,20 +5,21 @@ import store from './store'
...
-  })
+async function main () {
+  const session = await fetch('/api/sessions')
+  const data = await session.json()
+
+  if (data && data.user) {
+    await store.dispatch('login', data.user)
+  } else {
+    await store.dispatch('loadApplicationSettings')
+  }
```
igor47, line 16: this logic is still buggy. in particular, we only seem to set `viewPeriodId` in `loadApplicationSettings` and it ends up being `null` a lot of the time, which kinda still works because it should be set to the current period, and the backend generally also uses the current period if the specified period is not set, but this is not ideal. the entire flow of how the app loads it's initial state into the store should be re-worked.

src/server/services/persist-upload.js
```diff
@@ -26,7 +26,7 @@ async function extractDocuments (buffer) {
   for (const sheetName of workbook.SheetNames) {
     documents.push({
       type: normalizeSheetName(sheetName),
-      content: xlsx.utils.sheet_to_json(workbook.Sheets[sheetName], { header: 1 })
+      content: xlsx.utils.sheet_to_json(workbook.Sheets[sheetName], { header: 1, range: 'A1:AU200' })
```
igor47, line 29: i set this to avoid the problem discussed in slack, of having >1M rows in my spreadsheets. it might be nice to continue leaving this here? maybe we can limit it to something more sensible, like, say, 20k rows?

slack link: https://usdigitalresponse.slack.com/archives/C031R1Y49KL/p1651169022588859

src/store/index.js
```diff
@@ -14,6 +14,33 @@ export function get (url) {
   return fetch(url, options)
 }
 
+// this function always returns an object. in case of success, the object is
+// the JSON sent by the server. in case of any errors, the `error` property
+// contains a description of the error.
```
igor47, line 19: introducing a new pattern here, might be nice to update other server-interaction functions to also work like this.

src/store/index.js
```diff
@@ -151,6 +185,15 @@ export default new Vuex.Store({
         .map(r => (reportingPeriod.id === r.id ? reportingPeriod : r))
         .sortBy('start_date')
         .value()
+    },
+    updateAllUploads (state, updatedUploads) {
+      state.allUploads = updatedUploads
+    },
+    addAlert (state, alert) {
+      Vue.set(state.alerts, randomId(), alert)
```
igor47, line 193: i think `set` and `delete` below are vue-2 specific functions. in vue-3, i would declare the `alerts` object as `reactive` and then this would not be necessary.

src/main.js
igor47: referenced in #234

src/store/index.js
ajhyndman: Ooph.  I don't want to bog this PR down in a semantics debate, but I prefer the existing pattern.

Being able to use promise rejection to differentiate different failure modes is convenient.  That said, we aren't really investing heavily in error-handling at present!

i.e. You can do cool stuff like this

```
getJson('/api/endpoint')
  .catch(transportFailure => { return retry(); })
  .then(response => {
    if (!response.ok) {
      return Promise.reject();
    }
    // handle success
  })
  .catch(serverError => {
    // handle server error
  })
```

src/main.js
ajhyndman: Yikes, thanks for flagging this.

src/components/UploadHistory.vue
```diff
@@ -75,45 +82,15 @@ export default {
...
   watch: {
     periodId: async function () {
-      this.loadUploads()
+      this.$store.dispatch('updateUploads')
     }
   },
   mounted: async function () {
-    this.loadUploads()
+    this.$store.dispatch('updateUploads')
   }
```
ajhyndman, line 94: Thanks, this looks a lot more like I had in mind!

I fully support keeping the "series" and "documents" requests (in Uploads.vue) as local component state.

src/views/Upload.vue
```diff
@@ -219,61 +233,45 @@ export default {
...
+      } else {
+        this.series = result.series
+        this.seriesValid = result.currently_valid
       }
     },
     loadDocuments: async function () {
-      try {
-        const resp = await fetch(`/api/uploads/${this.uploadId}/documents`)
-        const result = (await resp.json()) || { error: (await resp.body) }
+      this.documents = []
```
ajhyndman, line 265: Question: Can we drop this.documents and the corresponding fetch from this component?

I think the extractDocuments code will be useful for building the treasury exports, but it doesn't look like it's rendered in the UI anywhere.

(This can definitely be a separate PR, and I'm happy to take a pass at it.)

src/views/Upload.vue
```diff
@@ -219,61 +233,45 @@ export default {
...
-            level: 'err'
-          }
-        }
-      } catch (e) {
-        this.alert = {
-          text: `loadUpload Unknown Error: ${e.message}`,
+      const result = await getJson(`/api/uploads/${this.uploadId}/series`)
+      if (result.error) {
+        this.$store.commit('addAlert', {
+          text: `loadSeries Error (${result.status}): ${result.error}`,
```
ajhyndman, line 256: Another possible future enhancement:

Is there detail in this request that isn't present in the `allUploads` store cache?  I think it'd be nice if this could be implemented as a `getter` on the store instead of triggering another server round trip.

ajhyndman (approved):
Okay!  I did have a bunch of thoughts while reviewing this, and left comments.

However, I think most of the items are discussion items for future decision-making.  Re-reading my noted, I don't see anything blocking!  👍

src/server/services/persist-upload.js
igor47: oh, interesting. i haven't tried that, i was assuming it would still be slow because it will attempt to iterate over the 1M rows to see if they're empty. the resulting data structure from this call does not have 1M json rows, so i think it's the attempt-to-access that's slow, but i'll double-check.

src/store/index.js
igor47: mmm, i guess i don't really like the promise syntax, i find promises to be hard to think about, and really prefer async/await syntax. guaranteeing that this function will always return an object helps simplify the call sites, since generally we just log the error to an alert modal. but i also like the example you have here, seems pretty concise. i'm not totally sure how to implement that in the `getJson` function.

src/views/Upload.vue
igor47: i have to double-check the DB code, we might not be consistent in how we load associations -- specifically, the user email address, and the agency code.

src/views/Upload.vue
igor47: yeah, i think i left it in from before i refactored this component, but we're not really using it. would be down with this as a separate PR.

src/store/index.js
igor47: i guess in python, i would declare a few exception classes, and throw relevant exceptions; i still find that easier to read than chained promise syntax, but maybe it's not very java-scripty? maybe mostly because you just always have one `catch` after `try` in js, and i guess you have to check `instanceof` inside the `catch` vs catching specific exceptions by type

src/server/services/persist-upload.js
ajhyndman: Oh, I see your point.  I forgot the read performance itself was the bottleneck.  Worth testing still, I guess.

src/store/index.js
ajhyndman: I think you can do something similar with async/await syntax.  The end goal is just to have a good mechanism to distinguish transport and server exception types.

And yeah, you're right. In python I'd definitely use exception subclasses!  I guess transport = reject / server = not ok just mimics the fetch API, so that's what I expect when I'm poking around a new codebase.

Again, we're not really handling transport failures anywhere else at all at the moment, so I don't feel that strongly.  😅

src/server/services/persist-upload.js
igor47: yeah, still a problem with `blankrows: false` --

```
2022-05-12T15:45:42.606Z persist-uploads extracting Certification (A1:D4
[dev:*express] 2022-05-12T15:45:42.606Z persist-uploads extracting Cover (A1:I13
[dev:*express] 2022-05-12T15:45:42.607Z persist-uploads extracting EC 1 - Public Health (A1:AR1048576
[dev:*express] 2022-05-12T15:46:14.961Z persist-uploads extracting EC 2 - Negative Economic Impact (A1:AU200
[dev:*express] 2022-05-12T15:46:14.966Z persist-uploads extracting EC 3 - Public Sector Capacity (A1:AJ200
```

ec1 ref is to 1048576 rows, and it takes 30 seconds

PR #238, opened 2022-05-11
Extract common helper functions for `log` and `dir`

### LINKED ISSUE #147: Extract VERBOSE logger function to util file
The helper function `log` is defined separately in many source files on the server.  We should just extract that to a single definition.

```js
let log = () => {}
if (process.env.VERBOSE) {
  log = console.log
}
```

ajhyndman:
Fixes #147

### DIFF
--- a/src/server/lib/log.js
+++ b/src/server/lib/log.js
@@ -1,0 +1,28 @@
+/**
+ * An alias for console.dir.
+ *
+ * This function is a no-op if process.env.VERBOSE is not set.
+ *
+ * @param {...*} messages
+ */
+let dir = (...messages) => {}
+if (process.env.VERBOSE) {
+  dir = console.dir
+}
+
+/**
+ * An alias for console.log.
+ *
+ * This function is a no-op if process.env.VERBOSE is not set.
+ *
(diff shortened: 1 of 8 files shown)


src/server/lib/log.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * An alias for console.dir.
+ *
+ * This function is a no-op if process.env.VERBOSE is not set.
+ *
+ * @param {...*} messages
+ */
+let dir = (...messages) => {}
+if (process.env.VERBOSE) {
```
igor47, line 9: rather than looking in `process.env` directly, can this setting live in `server/environment.js` ?

igor47 (approved):
lgtm, left a minor suggestion

i wonder if shouldn't even really be using `console.log`. express uses [debug](https://www.npmjs.com/package/debug) internally, and that's already in our bundle, i wonder if we should stick with it in our codebase too? it's nice to get timestamps for free.

src/server/lib/log.js
ajhyndman: Oh yeah.  Good idea!

PR #242, opened 2022-05-14
Update test:server with ARPA reporting periods

ajhyndman:
Port the general purpose server seed to the test:server scenario

This is useful, because I've been pointing my local instance at this server_test db, and now I can more easily reset my DB contents.

### DIFF
--- a/seeds/04_reporting_periods.js
+++ b/seeds/04_reporting_periods.js
@@ -1,9 +1,10 @@
 require('dotenv').config()
 
-/*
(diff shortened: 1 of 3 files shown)


igor47 (approved):
is this a straight-forward copy-paste from the seed in the root `seeds` directory? might be worth calling that out in both places

ajhyndman:
Yeah, this is almost a copy.  I think I recall there was a little bit of wrapping code that differed.  A link seems appropriate.  👍

PR #243, opened 2022-05-15
Fix close reporting period

ajhyndman:
This is my attempt to get the "Certify Reporting Period" button to work again.

I assume that the `crf_end_date` field doesn't mean anything for us?  I also don't know why we should need to preserve downloaded treasury reports on the server.


igor47 (changes requested):
looks good. can you also modify the non-test seed to remove the excess columns?

ajhyndman:
@igor47 I thought I had done that.  Is there another file I missed?

https://github.com/usdigitalresponse/arpa-reporter/pull/243/files#diff-413a00565138162a00e4c349d138f56d0973866fd9dc808098c0b4e247498542

PR #244, opened 2022-05-15
UPLOAD_DIR should be stable across restarts

ajhyndman:
@igor47 sorry, I didn't catch this in the initial PR.

Can you comment on your thinking when introducing this tmpdir approach for managing filesystem data?  I'm worried that using a randomly generated path in a path that the OS might clean up on some schedule isn't well-suited to the use case of persisting uploads.

At a minimum, doesn't this mean that redeploying the server will break references to previous uploads?

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,7 +13,6 @@ VALIDATION_TEMPLATE="validation-template.xlsx"
 # Agency input spreadsheet archives
 # This is /var/data/uploads in the states' deployed instances
 UPLOAD_DIRECTORY="./uploads"
-DATA_DIR="./datadir"
 
 COOKIE_SECRET=anystringwilldo
 INITIAL_ADMIN_EMAILS="admin1@example.com, admin2@example.com"
--- a/src/server/environment.js
+++ b/src/server/environment.js
@@ -1,17 +1,9 @@
 
-const { tmpdir } = require('os')
-const { mkdtempSync } = require('fs')
 const { join, resolve } = require('path')
-
-function tempDataDir () {
-  return mkdtempSync(join(tmpdir(), 'arpa-data-'))
-}
 
 const VERBOSE = Boolean(process.env.VERBOSE)
 
-const DATA_DIR = resolve(process.env.DATA_DIR || tempDataDir())
-const UPLOAD_DIR = join(DATA_DIR, 'uploads')
-const ARPA_REPORTS_DIR = join(DATA_DIR, 'arpa_reports')
+const UPLOAD_DIR = resolve(process.env.UPLOAD_DIRECTORY)
 
 const SRC_DIR = resolve(join(__dirname, '..'))
 const SERVER_DATA_DIR = join(SRC_DIR, 'server', 'data')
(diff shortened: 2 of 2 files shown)


src/server/environment.js
```diff
@@ -1,27 +1,17 @@
...
 
-function tempDataDir () {
-  return mkdtempSync(join(tmpdir(), 'arpa-data-'))
-}
-
 const VERBOSE = Boolean(process.env.VERBOSE)
 
-const DATA_DIR = resolve(process.env.DATA_DIR || tempDataDir())
-const UPLOAD_DIR = join(DATA_DIR, 'uploads')
-const ARPA_REPORTS_DIR = join(DATA_DIR, 'arpa_reports')
```
ajhyndman, line 14: Note that I am removing the only reference to ARPA_REPORTS_DIR here:

https://github.com/usdigitalresponse/arpa-reporter/pull/243/commits/765ce991dfe2e4e1116c12e118dfb528b1e0a24a#diff-b3759fc859de68384f315c38753e3036f1d4a875d873a490cab83e5a4cc091d6L152

.env.example
```diff
@@ -13,7 +13,6 @@ VALIDATION_TEMPLATE="validation-template.xlsx"
 # Agency input spreadsheet archives
 # This is /var/data/uploads in the states' deployed instances
 UPLOAD_DIRECTORY="./uploads"
-DATA_DIR="./datadir"
```
igor47, line 16: my goal was to have a single env var, which is `DATA_DIR`, vs lots of sub-env-vars like `UPLOAD_DIR` or `REPORT_DIR`

src/server/environment.js
```diff
@@ -1,27 +1,17 @@
...
 const { join, resolve } = require('path')
 
-function tempDataDir () {
-  return mkdtempSync(join(tmpdir(), 'arpa-data-'))
-}
-
 const VERBOSE = Boolean(process.env.VERBOSE)
 
-const DATA_DIR = resolve(process.env.DATA_DIR || tempDataDir())
-const UPLOAD_DIR = join(DATA_DIR, 'uploads')
```
igor47, line 13: i would love to keep this setting

src/server/environment.js
```diff
@@ -1,27 +1,17 @@
...
-const { mkdtempSync } = require('fs')
 const { join, resolve } = require('path')
 
-function tempDataDir () {
-  return mkdtempSync(join(tmpdir(), 'arpa-data-'))
-}
-
 const VERBOSE = Boolean(process.env.VERBOSE)
 
-const DATA_DIR = resolve(process.env.DATA_DIR || tempDataDir())
```
igor47, line 12: yeah, good question, why is this like that? fully support removing the temp functionality here.

igor47 (changes requested):
i would love to keep a single `DATA_DIR` directory with sub-paths by usage, as needed. fewer things to keep track of.

ajhyndman:
Ohh, this makes more sense.  Yeah, I can get behind making DATA_DIR the sole env variable.  We'll need to add it to the state render instances when landing this.

I've taken another pass at it!

igor47 (approved):

PR #245, opened 2022-05-15
Refactor some document and upload helpers

ajhyndman:
I think we want to be able to get all documents for a given period when generating a new export zipfile.  This PR reorganises a few helpers to support that.

### DIFF
--- a/src/server/db/uploads.js
+++ b/src/server/db/uploads.js
@@ -40,18 +40,6 @@ async function uploadsForAgency (agency_id, period_id) {
     .orderBy('created_at', 'desc')
 }
 
-async function uploadsForPeriod (period_id) {
-  console.log('uploadsForPeriod()')
-  if (!period_id) {
-    period_id = await getCurrentReportingPeriodID()
-  }
-
-  return knex('uploads')
-    .select('*')
-    .where({ reporting_period_id: period_id })
-    .orderBy('created_at', 'desc')
-}
-
 function upload (id) {
   return knex('uploads')
     .leftJoin('users', 'uploads.user_id', 'users.id')
@@ -160,7 +148,6 @@ module.exports = {
   upload,
   uploads,
   uploadsForAgency,
-  uploadsForPeriod,
   setAgencyId,
   setEcCode,
   markValidated,
--- a/src/server/db/uploads.js
+++ b/src/server/db/uploads.js
@@ -28,7 +28,6 @@ async function uploads (periodId, agencyId = null, onlyValidated = false) {
 }
 
 async function uploadsForAgency (agency_id, period_id) {
-  console.log('uploadsForAgency()')
   if (!period_id) {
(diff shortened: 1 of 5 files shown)


src/server/services/documents.js
```diff
@@ -0,0 +1,36 @@
...
+  debug('done extracting')
+  return documents
+}
+
+async function documentsForUpload (upload) {
+  return extractDocuments(await bufferForUpload(upload))
+}
+
+async function documentsForPeriod (periodId) {
+  const uploads = uploadsForPeriod(periodId)
```
igor47, line 29: you'll want to use the `validForReportingPeriod` function here, otherwise you'll just get all the uploads.

src/server/services/documents.js
```diff
@@ -0,0 +1,36 @@
+const debug = require('debug')('persist-uploads')
```
igor47, line 1: should probably be `documents`. we should also maybe namespace. if we do `require('debug')('arpa.documents')` then we can run `DEBUG='arpa.*' yarn serve` and get all of our debug output

igor47:
i'm assuming this is a draft/not ready for review, due to presence of debug logging and failing tests. can you re-request review if/when ready?

src/server/services/documents.js
ajhyndman: Oh cool.  That's a good idea.

ajhyndman:
> i'm assuming this is a draft/not ready for review, due to presence of debug logging and failing tests. can you re-request review if/when ready?

Oh, I definitely just copy-pasted the debug import without trying to comprehend what it was doing!  The failing test looks like just a lint oversight.  I'll fix both and push again.

src/server/db/uploads.js
```diff
@@ -28,8 +28,8 @@ async function uploads (periodId, agencyId = null, onlyValidated = false) {
 }
 
 async function uploadsForAgency (agency_id, period_id) {
+  console.log('uploadsForAgency()')
```
igor47, line 31: did you mean to leave this in?

src/server/db/uploads.js
```diff
@@ -40,6 +40,18 @@ async function uploadsForAgency (agency_id, period_id) {
     .orderBy('created_at', 'desc')
 }
 
+async function uploadsForPeriod (period_id) {
+  console.log('uploadsForPeriod()')
```
igor47, line 44: here too

src/server/db/uploads.js
```diff
@@ -40,6 +40,18 @@ async function uploadsForAgency (agency_id, period_id) {
     .orderBy('created_at', 'desc')
 }
 
+async function uploadsForPeriod (period_id) {
```
igor47, line 43: is this method still useful? doesn't look like it's getting called anywhere else in this PR

igor47 (approved):
left a few minor nits, seems good overall

PR #249, opened 2022-05-18
Additional UI for upload tracking

igor47:
This PR makes two user-facing changes. The first is replacing the `UploadHistory` component with an `Uploads` view, which lives in it's own tab. We kill displaying the upload history on the dashboard page. The dashboard page is unified, from two separate sub-components to just a single view file, and the space is made available for writing documentation or usage instructions.

Updated dashboard page:
image

The new `Uploads` view uses `vue-good-table` to make the uploads sortable, groupable, and filterable. Some screenshots in non-grouped and grouped mode:
image

image

Next, we add a UI for managing reporting-period wide validations. The UI looks like so:
image

to make this possible, we had to do a bunch of back-end refactoring. Most especially, we need to be more serious about transactions, since our dry-run mode for re-validation requires rolling back a transaction.


src/views/Home.vue
```diff
@@ -1,23 +1,82 @@
...
+        <a :href="downloadUrl()" class="btn btn-primary">Download Treasury Report</a>
+      </div>
+
+      <div class="col" v-if="isAdmin && false">
+        <a href="/api/audit_report" class="btn btn-info">Download Audit Report</a>
+      </div>
+
+      <div class="col">
+        <button @click.prevent="startUpload" class="btn btn-primary">Submit Spreadsheet</button>
+      </div>
```
ajhyndman, line 14:
```suggestion
      <div class="col" v-if="isAdmin">
        <a :href="downloadUrl()" class="btn btn-primary btn-block">Download Treasury Report</a>
      </div>

      <div class="col" v-if="isAdmin && false">
        <a href="/api/audit_report" class="btn btn-info btn-block">Download Audit Report</a>
      </div>

      <div class="col">
        <button @click.prevent="startUpload" class="btn btn-primary btn-block">Submit Spreadsheet</button>
      </div>
```

This is just a suggestion.  See screenshot above.

src/views/Validation.vue
```diff
@@ -0,0 +1,244 @@
...
+        </div>
+
+        <button class="btn ml-2" :class="revalidateBtnClass" @click="revalidate" :disabled="disableRevalidateBtn">
+          <div class="spinner-grow" role="status" v-if="revalidating">
+            <span class="sr-only">Loading...</span>
+          </div>
+
+          <span v-if="commit">Revalidate uploads</span>
+          <span v-else>Check upload validity</span>
+        </button>
```
ajhyndman, line 28:
```suggestion
        <button class="btn ml-2" :class="revalidateBtnClass" @click="revalidate" :disabled="disableRevalidateBtn">
          <span v-if="revalidating" class="spinner-grow spinner-grow-sm" role="status">
          </span>
          <span v-if="revalidating"> Loading...</span>
          <span v-else>
            <span v-if="commit">Revalidate uploads</span>
            <span v-else>Check upload validity</span>
          </span>
        </button>
```

I think the layout is a little wonky here.  maybe something like this is cleaner?

Before:

image

After:

image

src/server/routes/reporting-periods.js
```diff
@@ -194,6 +197,37 @@ router.get('/:id/template', requireUser, async (req, res, next) => {
...
+  }
+
+  const trns = await knex.transaction()
+  try {
+    const updates = await revalidateUploads(reportingPeriod, user, trns)
+    if (commit) {
+      trns.commit()
+    } else {
+      trns.rollback()
+    }
```
ajhyndman, line 219: I'm actually pretty happy with this solution for managing transactions!  Nice work.

src/server/db/users.js
```diff
@@ -117,32 +100,13 @@ function createAccessToken (email) {
...
-  knex,
   markAccessTokenUsed,
-  readSummaries,
   roles,
-  transact,
   updateUser,
-  upload,
-  uploads,
-  uploadsForAgency,
   user,
```
ajhyndman, line 110: These aliases were super weird.  Good call.  👍

src/server/environment.js
```diff
@@ -2,6 +2,8 @@ const { join, resolve } = require('path')
 
 const VERBOSE = Boolean(process.env.VERBOSE)
 
+const DB_URL = process.env.POSTGRES_URL
```
ajhyndman, line 5: Could we just make this `POSTGRES_URL`?  I don't see any reason for there to be two different names that developers have to remember match up.

src/views/Uploads.vue
```diff
@@ -0,0 +1,189 @@
...
+          filterOptions: {
+            enabled: true,
+            placeholder: 'Any agency',
+            filterDropdownItems: this.agencies.map(agency => ({ value: agency.code, text: agency.code }))
+          }
+        },
+        {
+          label: 'EC Code',
+          field: 'ec_code',
+          tdClass: (row) => { if (!row.ec_code) return 'table-danger' }
```
ajhyndman, line 127:
```suggestion
          tdClass: (row) => { if (!row.ec_code) return 'table-danger' },
          width: '120px'
```

Nit: On macOS, with Chrome and Firefox, this column appears to be rendered slightly too narrowly to fit in one row by default.

image

ajhyndman (changes requested):
This looks really good!

I'm pretty happy with both Vue Good Table and your transaction solution.

I did leave a few comments, but I think they are all in the layout polish category.

src/components/DownloadTemplateBtn.vue
```diff
@@ -0,0 +1,29 @@
...
+      return `/api/reporting_periods/${this.periodId}/template`
+    },
+    isDisabled: function () {
+      return !this.$store.getters.viewPeriodIsCurrent
+    },
+    classes: function () {
+      return {
+        'btn-secondary': this.isDisabled,
+        'btn-success': !this.isDisabled,
+        disabled: this.isDisabled
```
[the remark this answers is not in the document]
ajhyndman: Looks like I got a theme override to work here, for reference!

https://github.com/usdigitalresponse/arpa-reporter/pull/256

PR #250, opened 2022-05-18
Frozen input template

igor47:
we delete the un-used/stale validation template. we add the latest input template, as sent by joe.


igor47:
if we want to use the logic in the spreadsheet for validation, we can use the upload we're validating, which has all the validation logic in the header. we can also use the template for the current reporting period. i don't think the existing validation template is doing us any good.

my plan was to start implementing validations, and then see how it goes.

ajhyndman (approved):
I wasn't thinking about using the validation template for reading out actual validation logic.  It used to be that this validation spreadsheet was used as a template for validating that the right sheets were present, each with the right set of columns.

I guess that was already removed in a previous PR though, because I don't see it here.  You're also right, the information is redundant with info in the current reporting period's input template.  It's just a little harder to extract that info.  I'm okay with this change then.

PR #252, opened 2022-05-20
Purge all references and code relating to "documents"

ajhyndman:
This PR replaces "documents" with "records" in all remaining, useful code.

I was also able to remove large swaths of code that are no longer required with our streamlined records-based approach!


igor47 (approved):
i think we can probably kill even more code, but this is a good start.

PR #253, opened 2022-05-22
Omit empty CSV files from export

ajhyndman:
I also attempted a few performance optimizations here.

1. Fetch "records" for the reporting period only once, up front.
2. Don't bother fetching the template for a given CSV file until we know there is data.


igor47 (approved):
nice work, this is way cleaner

PR #255, opened 2022-05-22
Display a loading spinner while waiting for treasury export!

ajhyndman:
Shows a loading spinner and disables the treasury report download button while waiting for download.

Note that the algorithm for deciding when to consider the download complete is not great.  It turns out that browsers don't actually report any feedback when a download link has been resolved.  I found some crazy workarounds involving setting a cookie with the download response, but the effort to support that seems far beyond the value of the extra precision!

https://stackoverflow.com/questions/1106377/detect-when-a-browser-receives-a-file-download

Screen Shot 2022-05-21 at 11 32 11 pm

### DIFF
--- a/src/components/GrantsOfficeHome.vue
+++ b/src/components/GrantsOfficeHome.vue
@@ -8,8 +8,8 @@
         This reporting period is closed.
       </div>
       <div class="col-3" v-show="viewingCurrentPeriod">
-        <a href="/api/audit_report" class="btn btn-primary"
-          >Download Audit Report</a
+        <a href="/api/audit_report" class="btn btn-primary disabled"
+          >Download Audit Report 🚧</a
         >
       </div>
       <div class="col-3" v-show="viewingCurrentPeriod">
--- a/src/components/GrantsOfficeHome.vue
+++ b/src/components/GrantsOfficeHome.vue
@@ -2,7 +2,22 @@
   <div>
     <div class="row buttons mt-5">
       <div class="col-3">
-        <a :href="downloadUrl()" class="btn btn-primary">Download Treasury Report</a>
+        <a
+          :href="downloadUrl()"
+          class="btn btn-primary"
+          :class="{ disabled: isLoading }"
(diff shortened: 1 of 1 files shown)


src/components/GrantsOfficeHome.vue
```diff
@@ -46,6 +61,13 @@ export default {
   components: {
     UploadHistory
   },
+  mounted () {
+    window.addEventListener('focus', this.clearLoadingState.bind(this))
```
igor47, line 65: huh, how did you know that you need to re-bind `this`?

igor47 (approved):
this looks good, but it would be nice to re-open it against #249 which kills the `GrantsOfficeHome` component

src/components/GrantsOfficeHome.vue
ajhyndman: Ooph, JavaScript has some real rough edges when working with the `this` keyword.  I would say the reliance on `this` is my least favourite part of the Vue framework.

I don't even know where to start with an explanation!

I guess the main concept is that `this` in a JS function is generally determined by the context in which the function is _called_ and not the context in which it is defined.  That means that if a function has references to `this` and you intend to pass it as a callback somewhere else, then you can assume it those `this` references will be transferred to the calling code — unless you bind the current this value before passing.

This particular pattern shows up a lot in Vue (and in the class-style React components).  If you have a component method or lifecycle hook (which probably refers to this), but you plan to pass it as a callback somewhere, you need to rebind `this`, or it won't do what you expected.

There are some notes on the pattern in React docs, here:

https://reactjs.org/docs/handling-events.html

---

With all that said, I wrote it this way expecting it to be necessary, but I didn't actually test without.

I just poked at the VueJS docs, and at least in the latest version, it claims that the framework binds all functions in the `methods` object automatically, so it might not be required here!

https://github.com/vuejs/vue/blob/33e669b/src/core/instance/state.js#L286

ajhyndman:
Closing in favour of #258

PR #259, opened 2022-05-24
Multitenant: initial DB migration

mattbroussard:
As part of the "Preparation" section of the ARPA Reporter + GOST integration plan, we'll be adding a minimal version of multitenant support to the ARPA reporter tool. See integration plan doc: https://www.notion.so/usdr/GOST-ARPA-integration-technical-plan-cffc5800c0414660816a0920b959da82

This contains two DB migrations:
- **20220519161859_add_tenant_id.js**: adds the `tenant_id` field to various tables. The field specifies a default value of 0 which will be written to existing rows. Once all logic is updated to write a tenant_id explicitly, this default value can be removed from the column schema (in a separate migration) so that inserts lacking a tenant_id produce an error.
- **20220519182830_multitenant_unique_constraints.js**: updates `UNIQUE` and `FOREIGN KEY` constraints in affected tables so that things that were previously globally unique are now unique within each tenant.

I'm currently working on the next PR that begins to add filtering based on these fields; I'll probably hold off on merging this and running the migrations until I'm a bit farther into that, in case I realize additional DB schema changes that need to be made.


migrations/20220519161859_add_tenant_id.js
```diff
@@ -0,0 +1,51 @@
+const tables = [
```
ajhyndman, line 1: Nit: We've been using UPPER_SNAKE_CASE for module constants.

ajhyndman (commented):
The changes look great!  I appreciate all the thorough documentation.

If you wouldn't mind, it'd be great to satisfy the linter for now (even if we are considering moving to Prettier at some point).

---

In case it helps with anything, we're not actively using `period_summaries`, `subrecipients` or `projects` tables.

I believe @igor is in the process of rewriting `subrecipients`, and we will at some point iterate on `projects`, but I'm not sure we've committed to engaging with `period_summaries` yet.

mattbroussard:
Note to self: @igor47's #260 adds a new table `arpa_recipients` that will likely need a `tenant_id` field as well. Figure out which PR will land first and as such which needs to account for this.

mattbroussard:
> In case it helps with anything, we're not actively using period_summaries, subrecipients or projects tables.

Unless you're ready to remove code that references these tables, I think they still need to be dealt with here (and in my subsequent PRs), in case they start being more actively used.

E.g. for `period_summaries` there seems to be an endpoint `/api/reporting_periods/summaries/` that reads from this but that endpoint is not currently referenced anywhere on the client. But someone _could_ theoretically make requests to that endpoint.

igor47:
> they still need to be dealt with

yeah, we've inherited quite a lot of dead code in this project, and have been furiously deleting it, but there's still a lot of deletion to go

ajhyndman:
Yep, you got it working too, so no reason to backtrack!

I guess I'm just letting you know that if you find it easier to remove them at some point, that's probably fine too.

ajhyndman (approved):

ajhyndman:
Sorry, @mattbroussard were you waiting for us to land this change?

I'll go ahead and merge it.

mattbroussard:
No, I was waiting to see if I ran into things in the course of developing #282 that inspired other changes I'd want to make in this branch.

Also was waiting to see whether #260 landed first, in which case I'd update this PR to deal with its new table as well. Now that this has landed and #260 is still open, @igor47 you should probably just add a `tenant_id` field to your new table(s) at creation time.

ajhyndman:
Ahh, sorry!  Maybe I was too eager to break the stalemate.

Okay, I tried splitting apart #260 and added an extra commit to introduce the tenant_id column and constraints.

#296

PR #260, opened 2022-05-24
Igor/recipients

igor47:
We create a new `arpa_recipients` table. we prefix the table name with `arpa` to make this clear this is what this table does. We kill the old `subrecipients` table.

We populate this table when validating an upload. The validation logic is pretty complicated, and tries to handle cases where:
* an upload is re-submitted, and new validation rules persist the subrecipient differently
* a later upload also includes the same subrecipient, either identically (ok) or with differences (validation error)
* a subrecipient is modifed in the UI, and then the upload that first created it is re-validated

To deal with the validation, we begin parsing validation rules directly out of the spreadsheet -- either the upload, or the reporting period template.

We add a UI to display and search sub-recipients. We also add one to edit them. The edit UI relies on the validation rules to know about the possible set of fields in a sub-recipient record.


migrations/20220519070647_arpa_recipients_table.js
```diff
@@ -0,0 +1,17 @@
+exports.up = function (knex) {
+  return knex.schema
+    .createTable('arpa_recipients', function (table) {
```
ajhyndman, line 3: Are "recipients" here a different concept from "subrecipients" as defined in the export template?

I'm thinking in the UI we're probably going to want to be clear that these are "subrecipients", and it could be confusing if we call the same thing "subrecipient" one place and "recipient" somewhere else.

src/views/Subrecipient.vue
```diff
@@ -1,93 +1,145 @@
...
+          <span v-else>
+            By {{ recipient.updated_by_email }} on {{ humanDate(recipient.updated_at) }}
+          </span>
+        </div>
+      </div>
+
+      <div class="form-group row" v-for="(rule, key) in rules" :key="key">
+        <label :for="key" class="col-sm-2 col-form-label">{{ rule.humanColName }}</label>
+        <div class="col-sm-10">
+          <select v-if="rule.listVals.length > 0" :id="key" v-model="record[key]" :readonly="isReadOnly(key)">
```
ajhyndman, line 39:
```suggestion
          <select v-if="rule.listVals.length > 0" :id="key" v-model="record[key]" :readonly="isReadOnly(key)" class="form-control">
```

This should have the `form-control` class also.

ajhyndman (changes requested):
Okay, I think I want to propose breaking this PR up a little.

1. The ESLint rule is unrelated, we can easily land that in a quick PR first.
2. I'm happy with the DB schema changes, I don't see any reason not to land the migrations ASAP.
3. The Subrecipients table view looks great.

Subrecipient editing UI I want to talk more about.  Specifically, I don't like that we are diverging from RecordForm.  I think that's (a) more code to maintain, (b) stylistically inconsistent, and (c) it's not clear to me yet why it's necessary.

Relatedly, I haven't spent as much time as you looking at it, but I thought the validation DSL that we inherited from CARES was one of the nicest things about this repository.

https://github.com/usdigitalresponse/arpa-reporter/blob/staging/src/server/services/validate-data/subrecipients.js

If we lean more in the direction of compatibility with this code, does that make it easier to re-use RecordForm?

---

TL;DR:

- we can demo this tomorrow as-is
- I'd like to go through your rationale and maybe explore some more options before landing the validation and editing
- If you can break that out from the other changes in this PR, I'm happy to land them sooner

ajhyndman (approved):
Looks good to me!

Can you just verify that running the migrations works?  I suspect you might need to update the timestamp in the arpa_subrecipients migration filename in order to get it to run in the right order with Matt's migrations.

Issue #57, opened 2022-02-04
Export in ARPA-required formats


mhuang-usdr:
Expected export templates are here: https://drive.google.com/drive/u/0/folders/1Z-AS0acaY_HdO5Y4rbFNwBTOuxeMTsjm

Mapping from project expenditure category to templates starts on page 62: https://drive.google.com/file/d/1rNgA8bAUhUThJfzjKPJzKzYggOFp6dtt/view?usp=sharing

Supported templates:

- [x] expendituresGT50000BulkUpload.xlsx
- [x] expendituresLT50000BulkUpload.xlsx
- [x] paymentsIndividualsLT50000BulkUpload.xlsx
- [x] project18_229233BulkUploads.xlsx
- [x] project19_234BulkUploads.xlsx
- [x] project2128BulkUploads.xlsx
- [x] project214_224227BulkUploads.xlsx
- [x] project236BulkUploads.xlsx  
- [x] project31BulkUpload.xlsx
- [x] project32BulkUpload.xlsx
- [x] project4142BulkUpload.xlsx
- [x] project51518BulkUpload.xlsx
- [x] project519521BulkUpload.xlsx
- [x] projectBaselineBulkUpload.xlsx
- [x] subRecipientBulkUpload.xlsx
- [x] subawardBulkUpload.xlsx

ajhyndman:
I'm adding a checklist here to track progress!

- [x] expendituresGT50000BulkUpload.xlsx        
- [ ] expendituresLT50000BulkUpload.xlsx        
- [ ] paymentsIndividualsLT50000BulkUpload.xlsx 
- [ ] project18_229233BulkUploads.xlsx          
- [ ] project19_234BulkUploads.xlsx             
- [ ] project2128BulkUploads.xlsx               
- [ ] project214_224227BulkUploads.xlsx         
- [ ] project236BulkUploads.xlsx                
- [ ] project31BulkUpload.xlsx
- [ ] project32BulkUpload.xlsx
- [ ] project4142BulkUpload.xlsx
- [ ] project51518BulkUpload.xlsx
- [ ] project519521BulkUpload.xlsx
- [x] projectBaselineBulkUpload.xlsx
- [ ] subRecipientBulkUpload.xlsx
- [x] subawardBulkUpload.xlsx

ajhyndman:
It looks like EC code to template mapping starts on page 55 as of today (2022-05-24).

PR #269, opened 2022-05-25
Support exports: all EC 1 and EC 2 subcategories

ajhyndman:
This completes logic for the following functions:

- generateProject236
- generateProject214
- generateProject2128
- generateProject19
- generateProject18

In addition, I added mapped all EC 1 and EC 2 categories not covered by other templates to the project baseline template.

Addresses #57

### DIFF
--- a/src/server/services/generate-arpa-report.js
+++ b/src/server/services/generate-arpa-report.js
@@ -144,8 +144,36 @@ async function generateProject2128 (records) {
       case '2.6':
       case '2.7':
       case '2.8': {
-        // TODO
-        return null
+        return [
+          null, // first col is blank
+          record.type, // FIXME: transform from sheet tab to export format
+          record.subcategory,
+          record.content.Name,
+          record.content.Project_Identification_Number__c,
+          record.content.Completion_Status__c,
+          record.content.Adopted_Budget__c,
+          record.content.Total_Obligations__c,
+          record.content.Total_Expenditures__c,
+          record.content.Current_Period_Obligations__c,
+          record.content.Current_Period_Expenditures__c,
+          record.content.Does_Project_Include_Capital_Expenditure__c,
+          record.content.Total_Cost_Capital_Expenditure__c,
+          record.content.Type_of_Capital_Expenditure__c,
+          record.content.Type_of_Capital_Expenditure_Other__c,
+          record.content.Capital_Expenditure_Justification__c,
+          record.content.Project_Description__c,
+          record.content.Program_Income_Earned__c,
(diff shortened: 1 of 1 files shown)


src/server/services/generate-arpa-report.js
```diff
@@ -28,144 +54,422 @@ async function generateReportName (periodId) {
...
+        case '2.17':
+        case '2.18':
+        case '2.19':
+        case '2.20':
+        case '2.21':
+        case '2.22':
+        case '2.23':
+        case '2.28':
+        case '2.35':
+        case '2.37': {
```
ajhyndman, line 437: One thought:

It occurred to me that this list should logically be the inverse of any records matched by any other template.  It might be nice to think about rewriting the way we delegate records so that one record can only ever be assigned to a single export template.

However, I wasn't quite confident enough to tackle that now.

(a) It's not obvious to me what a simple algorithm that accomplishes this would be.
(b) Maybe that invariant won't hold (for example, what if we want automatically calculate project expenditure totals?)

src/server/services/generate-arpa-report.js
```diff
@@ -12,6 +11,33 @@ function isNotNull (value) {
...
+ */
+function getDetailedEcCode (record) {
+  const EC_CODE_REGEX = /^(\d.\d\d?)/
+  const { subcategory } = record
+
+  if (EC_CODE_REGEX.test(subcategory)) {
+    const [, detailedECCode] = EC_CODE_REGEX.exec(subcategory)
+    return detailedECCode
+  }
+  return undefined
```
igor47, line 38:
```suggestion
  const match = EC_CODE_REGEX.exec(subcategory)
  return match ? match[1] : undefined
```

avoids having to run the regex twice, plus seems a little easier to read

src/server/services/generate-arpa-report.js
igor47: in python i would do it with sets. so, for instance:
```python
all_codes = set(('1.1', '1.2', '2.5', '5.19', '5.20'))
code_to_output = {
   'generateProject519521': set('5.19', '5.20')
 }
all_assigned = reduce(lambda s1, s2: s1 | s2, code_to_output.values())
code_to_output['baseline'] = all_codes - all_assigned
```

igor47 (approved):
lgtm, left only the minorest of feedbacks

src/server/services/generate-arpa-report.js
ajhyndman: 👌   I considered a few approaches to this, and now I know which you prefer!

PR #271, opened 2022-05-25
Set "engines" in package.json

### LINKED ISSUE #64: set/maybe upgrade the nodejs version
the node version to use with the project is unclear. we should add guidance about that to the readme and maybe a file used by nodenv to make it more clear. we might want to upgrade the nodejs version. we should coordinate the upgrade with the various render instances that run this code -- there might be render configuration that specifies the version.

ajhyndman:
Low priority!

I opted to set a node version via the "engines" property of our `package.json`.  The advantage of setting a version this way is that yarn will also be able to enforce it locally.

This ensures that the version of node we run locally won't diverge significantly from what's running in production.

I chose v16 since v16 is the latest version of node that meets the following criteria:

- in "long term support" ([LTS](https://nodejs.org/en/about/releases/))
- supports our version of Vue CLI
- supports our version of Express
- was released at least 6 months ago (node 18 was released 2022-04-19)

I didn't identify any breakages on v18, so we could try it if we wanted.  v16 just seemed a smidge safer.

Fixes #64

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
   "name": "arpa-reporter",
   "version": "0.1.0",
   "private": true,
+  "engines": {
+    "node": "16.x.x"
+  },
   "scripts": {
     "serve": "concurrently \"npm:dev:*\"",
     "dev:vue": "vue-cli-service serve",


mattbroussard:
This is a newer version than the GOST project is currently at. It might make the integration into GOST simpler to set it the same for now (GOST specifies 14.19.0 via `.node-version`/`.nvmrc`), otherwise part of the integration project will have to be bringing GOST's Node version forward.

Upgrading Node in GOST would be good to do anyway -- I remember when I did my initial dev setup a few weeks ago, the repo was not automatically compatible with Node 17 that was on my system and I needed to build Node 14 from source (a configuration not officially supported on Apple Silicon but "experimental" and seems to work ok). It seems to be because of incompatibilities in `node-sass` and `@vue/cli` (see https://github.com/vuejs/vue-cli/issues/6853) -- both have newer versions now that support Node 16 but not sure if there are other things depending on those that would complicate the upgrade.

There is a task #45 (and experimental [branch](https://github.com/usdigitalresponse/usdr-gost/tree/45-upgrade-app-to-vuejs) from @Rapol) about trying to upgrade to Node 16 and Vue 3 in and a decision apparently made on March 23 to punt on the upgrade. @Rapol can you add any historical context about this decision & how difficult it would be to e.g. upgrade GOST to Node 16 without necessarily changing Vue major version?

ajhyndman:
Oh, thanks for the note!

Node 14 should handle everything we're doing here — in fact that's probably what's running on render instances — so I'm happy to drop back down for now.

ajhyndman:
It's still fine to drop down to v14 if we prefer, but I was curious what the bottleneck in usdr-gost actually was (we are using Vue v2 in the arpa project as well).

I think I might have worked it out:

https://github.com/usdigitalresponse/usdr-gost/pull/83

If it makes life easier for any future apple silicon devs, it could be worth moving toward node 16.

ajhyndman:
Closing in favour of #366

PR #273, opened 2022-05-25
Bring `main` up to `staging` for milestone 1

### LINKED ISSUE #266: Revise auto-deployment processes
After May 26th we will have active users.  We should reconsider our automatic deployment process.

I propose considering something like the following:

- Keep the USDR staging deployment pointed to the `staging` branch
- Reset the `main` branch to latest commit of `staging`
- Point RI and OH branches to `main` branch
- Manually advance `main` after communicating new set of changes to users

igor47:
closes #266 

i would wait until we have #260 merged into staging to deploy this and do the operational steps. the steps are:
* re-point RI and OH to the `main` branch
* run migrations from render console


ajhyndman:
I think it might actually be safer to just `reset` main straight to the current commit from staging.

Using a merge, we aren't 100% guaranteed that there won't be some difference between `main` and `staging`, are we?

I'm imagining something like this.

```
$ git checkout main
$ git reset origin/staging --hard
$ git push --force
```

PR #275, opened 2022-05-26
Initialize and execute regex only once

ajhyndman:
Addresses some feedback from #269.

---

While we're talking regular expressions, I found this library the other day.

https://github.com/jribbens/verbose-regexp

I believe it's inspired by the verbose-regex feature in python, and it seems like a pretty cool way to get at the power of regular expressions without having to write dense, un-documentable strings of gibberish.

Maybe that's worth trying to incorporate sometime?

### DIFF
--- a/src/server/services/generate-arpa-report.js
+++ b/src/server/services/generate-arpa-report.js
@@ -6,6 +6,20 @@ const { applicationSettings } = require('../db/settings')
 const { getTemplate } = require('./get-template')
 const { recordsForReportingPeriod } = require('./records')
 
+const EC_CODE_REGEX = /^(\d.\d\d?)/
+
+/**
+ * Extract the Detailed Expenditure Category code from a record.
+ *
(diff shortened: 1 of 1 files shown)


src/server/services/generate-arpa-report.js
```diff
@@ -6,6 +6,20 @@ const { applicationSettings } = require('../db/settings')
...
+/**
+ * Extract the Detailed Expenditure Category code from a record.
+ *
+ * @returns {string} The detailed EC code in format "#.##".
+ */
+function getDetailedEcCode (record) {
+  const { subcategory } = record
+
+  const match = EC_CODE_REGEX.exec(subcategory)
+  return match?.[1]
```
ajhyndman, line 20: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#syntax

igor47 (approved):

PR #277, opened 2022-05-26
Use custom bootstrap file input

ajhyndman:
Install the bootstrap custom file input plugin and apply it to our file input.

Fancier, but we load a few extra kilobytes, and there's a chance there are weird bugs.

Screen Shot 2022-05-25 at 7 13 44 pm

Alternative: #278


igor47:
i prefer the simpler alternative. if we do add a file upload element, we should pick one that allows selection of multiple files, and then maybe put a multi-file picker in the admin UI (this was requested verbally in our last demo meeting)

PR #280, opened 2022-05-26
Upload: correct display of in-series uploads

### LINKED ISSUE #276: Not selecting the most recent valid upload for exports
It doesn't appear that the ARPA tool is always correctly identifying the most recently validated upload for use in exports.

image

igor47:
we had several logic bugs here. first, our series display was listing
uploads for other EC codes, since we were only filtering by agency.
next, we likewise were also filtering only by agency when determining
the current valid period.

we refactor our db functions to use the same base query for consistency.
we also better-name our db query functions after what they're used for,
removing the overly-general `listUploads` (the filtering by various
fields there has moved to the front end in vue-good-table)

fixes #276

### DIFF
--- a/src/server/db/uploads.js
+++ b/src/server/db/uploads.js
@@ -5,37 +5,28 @@ const {
   getCurrentReportingPeriodID
 } = require('./settings')
 
-async function listUploads ({ periodId, agencyId = null, onlyValidated = false }, trns = knex) {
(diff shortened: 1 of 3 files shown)


ajhyndman (approved):
Awesome 👍 

I have a local repro and can confirm that this resolves the issue!

PR #282, opened 2022-05-26
Multitenant: Add filtering by tenant_id

mattbroussard:
Depends on #259 

See overall GOST integration plan: https://www.notion.so/usdr/GOST-ARPA-integration-technical-plan-cffc5800c0414660816a0920b959da82#b4dfb0449fc54674ae97f070a6b7b33a

This PR adds filtering and access checks based on the current user's `tenant_id` in (as far as I know) all places that need it. In most places the tenant ID to filter by comes from the logged in user, via a new `req.session.user` object populated by `requireUser` middleware that's used almost everywhere. In a few cases, functions that deal with other types of entities (mainly uploads) get a tenant ID from that object.

Most changes are serverside only and clients are generally not aware of the concept of tenants. My approach was to go through all the DB accessor functions in src/server/db and trace backward to the places that call them. This was a bit tricky in the absence of TypeScript since I had to rely on grep and we do some annoying things like renaming exports and imports.

One notable clientside change is that previously `/api/application_settings` was called before logging in and thus could not use `requireUser` to determine a tenant ID to filter by. I've changed this so that this data is not loaded until after login now, and the endpoint requires authentication. Accordingly, the navbar with the period dropdown (which draws from this data) is now hidden before login. There is a slight race condition after logging in since nothing explicitly waits for `application_settings` to be populated, but the only apparent issue in the gap before it loads is that the period selector is not populated. This doesn't seem like a huge deal, and that gap is very short.

Regarding dead code:
 - This PR removes `routes/files` and all code referencing `UPLOAD_DIRECTORY`. This seemed to be dead (posted about it on Slack [here](https://usdigitalresponse.slack.com/archives/C031R1Y49KL/p1653610770392909)) and was problematic to keep around because it doesn't interact with the database and couldn't be easily tenant-scoped. Also, the implementation of the /api/files/delete endpoint was insecure and could have been used to arbitrarily delete any file on the machine!
 - <del>I did not attempt for now to add filtering logic in `db/subrecipients` or `routes/subrecipients` since there is a concurrent PR https://github.com/usdigitalresponse/arpa-reporter/pull/260 that is removing/revamping these. I had previously made changes to several other now-deleted files that caused some issues when rebasing over https://github.com/usdigitalresponse/arpa-reporter/pull/249</del> edit: now that #260 has merged, I did go back and add tenantId checks to code dealing with the revamped subrecipients model.

For DB seeds, the "dev" ones (which I think may also be used in prod deployments? Staging at least has data that looks like it came from them) still assume a single tenant with ID `0`. Some of the "test" seeds add data for multiple tenants so that the tests can verify multitenant behaviors.

In this PR I also added a mechanism for writing API route tests using [Supertest](https://github.com/visionmedia/supertest) (this is conceptually similar to how API route tests work in GOST but imo a bit cleaner and something we might want to move GOST over to eventually). I added a few route tests verifying this works (health & sessions) and a few (exports) for multitenant, and I'd like to add some more, but I'm going to try to put those in a separate PR so this one stops growing and can land sooner.


src/server/db/reporting-periods.js
```diff
@@ -111,7 +119,8 @@ async function isCurrent (periodID) {
 /* closeReportingPeriod()
   */
 async function closeReportingPeriod (user, period, trns = knex) {
-  const reporting_period_id = await getCurrentReportingPeriodID()
+  // TODO(mbroussard): should this call pass trns?
```
ajhyndman, line 122: Yes, passing transaction here seems beneficial.

src/server/db/reporting-periods.js
```diff
@@ -146,25 +155,31 @@ async function closeReportingPeriod (user, period, trns = knex) {
     .where({ id: reporting_period_id })
     .update({
       certified_at: new Date().toISOString(),
+      // TODO(mbroussard): is this correct? We write a stringified json of user row?
```
ajhyndman, line 158: That does seem weird.  I can open a ticket and look into it separately.

src/server/db/subrecipients.js
```diff
@@ -104,7 +108,11 @@ function subrecipientById (id, trns = knex) {
 }
 
 async function createSubrecipient (subrecipient, trns = knex) {
-  subrecipient.created_in_period = await getCurrentReportingPeriodID()
+  // TODO(mbroussard): replace with actual plumbing of tenantId field
+  const tenantId = 0;
+
+  // TODO(mbroussard): should this pass trns?
```
ajhyndman, line 114: i think passing trns is more correct, yes.

src/server/db/uploads.js
```diff
@@ -28,8 +32,14 @@ async function listUploads ({ periodId, agencyId = null, onlyValidated = false }
...
 
 async function uploadsForAgency (agency_id, period_id, trns = knex) {
+  const agency = agencyById(agency_id, trns);
+  if (!agency) {
+    throw new Error('invalid agency in uploadsForAgency');
+  }
+
   if (!period_id) {
-    period_id = await getCurrentReportingPeriodID()
+    // TODO(mbroussard): should this pass trns?
```
ajhyndman, line 41: 👍

ajhyndman (commented):
Directionally, this looks great.  Feel free to leave some of these TODOs in the code for @igor47 and I to resolve.

Here are my thoughts on how to adapt seeds:

We currently use db seed files primarily for unit tests and local execution.  These shouldn't run in any production or staging environments.  From that point of view, whatever changes we make to seed files should align to the goal of giving developers a basically usable configuration after running.

One approach could be to set initial users, reporting periods and application settings to the default tenant_id.

mattbroussard:
Github seems to have done some confusing thing with my stacked PRs... Since #259 was merged, it says:
> Base automatically changed from multitenant-migrations-1 to staging [13 days ago](https://github.com/usdigitalresponse/arpa-reporter/pull/282#event-6695570896)

But then there are commits I've pushed to this branch past `multitenant-migrations-1` that are not in `staging` yet, but they do not show up on the PR page. The changes from them do show up in "Files changed" though.

Specifically:
- 6faa25d chore: rm TODO comment about missing user arg; the bad callsite was deleted
- 9f4482b chore: add missing trns to a few getCurrentReportingPeriodID callsites

src/server/db/reporting-periods.js
mattbroussard: #332 removed this.

src/server/routes/agencies.js
```diff
@@ -33,10 +34,19 @@ router.post('/', requireAdminUser, async function (req, res, next) {
 
   try {
     if (agencyInfo.id) {
+      const existingAgency = await agencyById(agencyInfo.id)
+      if (!existingAgency || existingAgency.tenant_id !== req.session.user.tenant_id) {
```
igor47, line 38: huh, how would the second condition ever happen?

src/server/routes/exports.js
```diff
@@ -4,11 +4,17 @@ const _ = require('lodash')
```
igor47, line 4: don't think the exports routes are in-use

src/server/routes/configuration.js
```diff
@@ -0,0 +1,17 @@
+/* eslint camelcase: 0 */
```
igor47, line 1: was this route already removed in staging? why is it being re-added?

src/server/routes/uploads.js
```diff
@@ -31,11 +31,9 @@ router.post('/', requireUser, multerUpload.single('spreadsheet'), async (req, re
     console.log('Filename:', req.file.originalname, 'size:', req.file.size)
   }
 
-  const user = await getUser(req.signedCookies.userId)
```
igor47, line 34: did you kill every instance of this pattern in the controllers? it's been pretty common, and it would be nice to consistently switch to just `const user = req.session.user`

igor47 (approved):
left comments of varying pertinence throughout the diff here, but nothing is blocking. maybe just go through my comments and make sure nothing jumps out at you?

src/components/Navigation.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="container-fluid" style="width: 90%">
-    <nav class="row navbar navbar-expand navbar-light bg-light">
+    <nav class="row navbar navbar-expand navbar-light bg-light" v-if="loggedIn">
```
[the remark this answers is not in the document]
mattbroussard: I talked about this in the PR description: this hides the reporting period selector when you're on the login page. It doesn't make any sense there anyhow, but now it won't have any data because `updateApplicationSettings` isn't run until after login (because `/api/application_settings` now requires auth, because it needs to know what tenant to lookup settings for; on the login page we definitionally _don't know_ which tenant you'll login to yet)

Though I'm realizing now maybe you mean why hide the whole navbar vs just the period selector, and true, I could change that. Though this also pulls `applicationTitle` from `application_settings` which has the same problem mentioned above, though there is already an explicit fallback to "ARPA Reporter" -- in this case, prob would want to change this to just be a fixed title and not use that field in `application_settings` anymore?

src/main.js
```diff
@@ -11,8 +11,6 @@ async function main () {
 
   if (data && data.user) {
     await store.dispatch('login', data.user)
-  } else {
-    await store.dispatch('updateApplicationSettings')
```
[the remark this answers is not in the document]
mattbroussard: (Answered above and in PR description)

src/server/db/agencies.js
```diff
@@ -6,8 +6,13 @@ function baseQuery (trns) {
     .select('*')
 }
 
-function agencies (trns = knex) {
+function agencies (tenantId, trns = knex) {
+  if (tenantId === undefined) {
```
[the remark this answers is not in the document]
mattbroussard: Because the default tenant ID is `0` which is falsey, so `if (tenantId)` is not enough.

src/server/services/validation-rules.js
```diff
@@ -93,9 +93,16 @@ async function rulesForUpload (upload) {
   return extractRules(await bufferForUpload(upload))
 }
 
-async function rulesForPeriod (periodId) {
+async function rulesForPeriod (tenantId, periodId) {
```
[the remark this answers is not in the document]
mattbroussard: It's removed in #348 which I just now merged against.

src/server/routes/uploads.js
mattbroussard: I think I got them all?

src/server/routes/exports.js
mattbroussard: Well, the code exists still and it theoretically could be called, so it needs to obey multitenant rules. If you want to kill it after this PR merges, that's cool 👍

src/server/routes/configuration.js
mattbroussard: Oh good catch, this happened accidentally while resolving merge conflicts. Will remove.

src/server/routes/agencies.js
mattbroussard: Someone could do a request passing in an agency ID that is valid but refers to an agency of a different tenant.

src/server/index.js
```diff
@@ -1,4 +1,9 @@
 const express = require('express')
+
+// This patches Express to better handle errors thrown in async functions, which are otherwise not
+// handled. It just calls next(err) with them.
```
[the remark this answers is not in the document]
mattbroussard: Not handled means:
 - Express's `next` (or `res.send` etc.) function never gets called, and the request is [left hanging](https://expressjs.com/en/guide/writing-middleware.html) until presumably it times out (and then maybe Express sends a 500 or something, I don't remember)
 - The promise rejection will trigger Node's `unhandledRejection` event, which we also do not handle in our app. It just prints a warning to stderr by default.

Express 5.0 natively handles errors in async middleware/route handlers so a plugin like `express-async-errors` is unnecessary, but 5.0 is still listed as beta so I don't suggest updating to it.

seeds/06_migrate_from_source.js
```diff
@@ -1,22 +1,27 @@
 require('dotenv').config()
 
-exports.seed = async function (knex) {
+// NOTE(mbroussard): unclear to me if this seed will be used again, but I'm updating it to minimally
```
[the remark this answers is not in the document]
mattbroussard: OK. I'm probably not going to do that in this PR, but if you'd like to kill it separately, go ahead 👍

seeds/04_reporting_periods.js
```diff
@@ -1,14 +1,22 @@
...
+// created it will need reporting periods created for it.
+
 // reporting periods loosely based on tables 3 and 4 from here:
 // https://home.treasury.gov/system/files/136/SLFRF-Compliance-and-Reporting-Guidance.pdf
 //
 // when making changes to this file, consider also updating the test seed:
 // tests/server/seeds/07_reporting_periods.js
-exports.seed = async function (knex) {
-  const [{ count }] = await knex('reporting_periods').count('name', { as: 'count' })
+exports.seed = async function (knex, tenantId = 0) {
```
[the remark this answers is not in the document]
mattbroussard: It won't get run with any other param by knex natively, but I parameterized this so it's easy to use for populating a new tenant if there's a need to do that in the future. You could either manually hardcode a different value in this one place, or make a different seed file that imports this one and calls with a parameter (potentially multiple times if needed).

I could rewrite this as something like this if you want. It would be slightly safer if e.g. a future version of knex passes multiple arguments of its own.
```
async function seedReportingPeriods(knex, tenantId = 0) {
  // ...
}
exports.seed = knex => seedReportingPeriods(knex)
```

tests/server/routes/exports.spec.js
```diff
@@ -0,0 +1,52 @@
+
+const { makeTestServer, getSessionCookie } = require('./route_test_helpers')
+
+describe('/api/exports', () => {
```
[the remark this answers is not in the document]
mattbroussard: As [mentioned](https://github.com/usdigitalresponse/arpa-reporter/pull/282#discussion_r906418773) I'm not going to do this now, but if you remove `exports` obviously you can remove this test at the same time.

src/components/Navigation.vue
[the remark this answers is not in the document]
mattbroussard: Oh, actually, it seems like you added an equivalent check in 4f08dcb7be3d41174a8d7c645489038f9e230fc8 after I'd done this in my branch, so it's redundant now.

src/views/Validation.vue
```diff
@@ -200,6 +200,9 @@ export default {
     },
     periodId: function () {
       return this.$store.state.viewPeriodID
+    },
+    periodName: function () {
+      return this.$store.getters.viewPeriod.name
```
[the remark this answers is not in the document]
mattbroussard: Think I'm going to leave this as-is for now since `periodId` is referenced in a watcher

src/server/routes/exports.js
mattbroussard: Hmm, it does seem in use: the main "Download Treasury Report" button on the dashboard links to `/api/exports`

PR #294, opened 2022-05-26
Add tsconfig and @typescript-eslint/no-floating-promises lint rule

mattbroussard:
As discussed in Slack, forgetting to `await` a Promise is a frequent pain point. This adds a TypeScript-based lint rule that picks this up and gives you errors like this:
```
/Users/mattb/Desktop/usdr/arpa-reporter/src/server/lib/audit-report.js
  590:5  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises

/Users/mattb/Desktop/usdr/arpa-reporter/src/server/lib/subrecipients.js
  44:9  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises

/Users/mattb/Desktop/usdr/arpa-reporter/src/server/routes/application_settings.js
  9:3  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises

/Users/mattb/Desktop/usdr/arpa-reporter/src/server/services/validate-upload.js
  118:5  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
  122:5  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises

✖ 5 problems (5 errors, 0 warnings)
```
These existing errors are also fixed in this PR.

Though we are not using TS yet, TS supports some basic typechecking of vanilla JS files based on type inference, JSDoc comments, and "implicit any" (there are some important [limitations](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html)). Right now running `tsc` produces 23 errors that are mostly due to the lack of explicit type definitions, but I don't plan to spend time right now fixing these or actually converting things to TS; I just added TS so that we could use the TS-based lint rule. Also, I've currently scoped this rule and the tsconfig file only to the `src/server` folder to keep things simple, but if we move to TS in the future, we'll want to expand that to also typecheck tests, Vue components/other clientside code, etc.

Note that there is also a non-TypeScript lint rule, [`no-floating-promise`](https://github.com/SebastienGllmt/eslint-plugin-no-floating-promise) that tries to enforce this in a much weaker way (i.e. only in a single file and with less type awareness) -- this TS version is superior since it evaluates type information across multiple files.

### DIFF
--- a/src/server/routes/application_settings.js
+++ b/src/server/routes/application_settings.js
@@ -5,10 +5,9 @@ const express = require('express')
 const router = express.Router()
 const { applicationSettings } = require('../db/settings')
 
-router.get('/', function (req, res) {
-  applicationSettings().then(application_settings =>
-    res.json({ application_settings })
-  )
+router.get('/', async function (req, res) {
+  const application_settings = await applicationSettings()
+  res.json({ application_settings })
 })
 
 module.exports = router
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,112 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig.json to read more about this file */
+    // NOTE(mbroussard): the commented out entries are *examples*, not defaults
+
+    /* Projects */
+    // "incremental": true,                              /* Enable incremental compilation */
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    // "tsBuildInfoFile": "./",                          /* Specify the folder for .tsbuildinfo incremental compilation files. */
+    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects */
+    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
+    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
+
+    /* Language and Environment */
+    "target": "es2015",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
+    // "lib": [],                                        /* Specify a set of bundled library declaration files that describe the target runtime environment. */
+    // "jsx": "preserve",                                /* Specify what JSX code is generated. */
+    // "experimentalDecorators": true,                   /* Enable experimental support for TC39 stage 2 draft decorators. */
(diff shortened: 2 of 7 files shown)


src/server/routes/application_settings.js
```diff
@@ -5,10 +5,9 @@ const express = require('express')
 const router = express.Router()
 const { applicationSettings } = require('../db/settings')
 
-router.get('/', function (req, res) {
-  applicationSettings().then(application_settings =>
-    res.json({ application_settings })
-  )
+router.get('/', async function (req, res) {
```
mattbroussard, line 8: Note that an error in this `async` function is technically unhandled (in a way that the new linter will not catch) since Express does not properly support async functions.

In my other PR #282 I ended up adding a third party module `express-async-errors` that addresses this as a sort of drive-by fix; though maybe that would fit better in this PR or separately-- it just happened to be useful while I was debugging my changes.

ajhyndman:
Nice, you got it working!

@igor47 I'd be curious if you can get this change to play nicely with your IDE, and if you find it helpful.

src/server/routes/application_settings.js
ajhyndman: It sounds like express 5.0 fixed this.  I wonder how big of an upgrade that is.

https://expressjs.com/en/guide/migrating-5.html#rejected-promises

mattbroussard: Express 5 is still listed as beta, so I don't think we should really consider switching to it yet.

ajhyndman: Ah, you're right.  Well, no rush.

tsconfig.json
```diff
@@ -0,0 +1,112 @@
+{
```
ajhyndman, line 1: I added a jsconfig.json file to the root of the repo to enable some VS Code in-IDE typechecking.

That should be redundant with this tsconfig file.  Could you remove that in this PR?

igor47 (approved):
this works for me. i rebased to fix the merge conflict, and removed the jsconfig.json file as requested by andrew.

PR #296, opened 2022-05-27
subrecipients [1] — New DB model

ajhyndman:
This is the first half of #260.

In the interest of unblocking progress I tried breaking this PR into two parts.

---

I added one commit to the end that adds a tenant_id and composite uniqueness constraints to the new table.


igor47:
closing in favor of #260 -- i cherry-picked the commit to add `tenant_id` to that PR

PR #297, opened 2022-05-27
subrecipients [2] — UI changes

ajhyndman:
This is the second half of https://github.com/usdigitalresponse/arpa-reporter/pull/260.

In the interest of unblocking progress I tried breaking this PR into two parts.

PR #308, opened 2022-06-04
Modularize VueX store

mattbroussard:
This is part of the project to integrate into the GOST repo. See overall GOST integration plan: https://www.notion.so/usdr/GOST-ARPA-integration-technical-plan-cffc5800c0414660816a0920b959da82#b4dfb0449fc54674ae97f070a6b7b33a

In GOST, different parts of the VueX [store](https://github.com/usdigitalresponse/usdr-gost/blob/main/packages/client/src/store/index.js) are separated into [modules](https://vuex.vuejs.org/guide/modules.html). This rearranges ARPA Reporter's store to mirror that structure. This may not be immediately necessary if we keep them as separate Vue pages in the same repo, but will eventually be needed if we merge them into a single unified Vue app/page. It may also make it easier to swap in modules from GOST even in the multi-page setup, such as `users` and `agencies` -- though it's not perfectly seamless since their structures + actions are slightly different.

This looks like a big PR but mostly it's just code moving from store/index.js to separate files. The things to look at are references to state fields, getters, mutations, and actions and be sure I found all the places to update -- and that the changes do not have any implications for Vue reactivity (I'm still not totally clear on how this works under the hood for computed fields that reference VueX state fields)

Some references are a bit uglier now since they're nested under another level of structure (e.g. `this.$store.getters['users/loggedInUser'].role` or `this.$store.stage.agencies.agencies`) but these can be later cleaned up by adopting the [`mapGetters`](https://vuex.vuejs.org/guide/getters.html#the-mapgetters-helper) (& friends like `mapMutations` and `mapActions`) pattern already used in GOST.

Note: I intentionally did not modularize the store fields + actions related to subrecipients here since #296 and #297 significantly change code that references these.


src/store/application_settings.js
```diff
@@ -0,0 +1,34 @@
...
+    title: null
+  },
+  mutations: {
+    setApplicationSettings (state, applicationSettings) {
+      Object.assign(state, applicationSettings)
+    }
+  },
+  actions: {
+    reloadApplicationSettings ({ commit }) {
+      return fetch('/api/application_settings')
```
mattbroussard, line 15: Note to self: need to add `{credentials: 'include'}`

igor47: > Note to self: need to add `{credentials: 'include'}`

i don't think this is actually true; i've been omitting it and things work fine, and i'm not really sure why it was added in the first place. we're not making cross-origin requests, and i *think* browsers default to sending them to same-origin sites (e.g., that the default is `same-origin`)

mattbroussard: Ah. I think you're right as long as these requests are all relative on the same host. In GOST dev mode, API_DOMAIN and WEBSITE_DOMAIN differ and the client served on localhost:8080 makes AJAX requests to localhost:3000, which I think is considered cross-origin? So that's probably where this came from originally (or rather, some common ancestor that had this behavior)

It seems annoying that GOST does that though and it would be easy to make everything be on the same origin in dev by setting
```
proxy: {
  '/api': {
    target: 'http://localhost:3000'
  }
}
```
in GOST's vue.config.js like ARPA Reporter has.

mattbroussard:
From [Slack](https://usdigitalresponse.slack.com/archives/C031R1Y49KL/p1656017529071409):
>Re: the VueX PR (https://github.com/usdigitalresponse/arpa-reporter/pull/308) I think unfortunately it's going to have to be abandoned and re-started; the merge conflicts are just pretty severe and I don't have confidence in not introducing issues in the course of trying to resolve them (since most of that diff is copying code from one big file to smaller ones, then changing a bunch of references). I think it'll be easier to just start over on it. I might get something up for that before I leave but very unlikely to merge it before I leave. This is not immediately blocking to the 2nd phases of the integration project since we decided to start that with ARPA app mounted on a separate page, so it can have a completely separate VueX store.

PR #311, opened 2022-06-09
feat: uploads table shows only exported uploads

### LINKED ISSUE #285: Filter uploads view for uploads that will be used

igor47:
we add a route to pull those from the reporting period. we update the db
query to be consistent with other upload db queries.

fixes #285

### DIFF
--- a/src/server/db/uploads.js
+++ b/src/server/db/uploads.js
@@ -41,11 +41,10 @@ function getUpload (id, trns = knex) {
 }
 
 function validForReportingPeriod (period_id, trns = knex) {
-  return trns.with('agency_max_val', trns.raw(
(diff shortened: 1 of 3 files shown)


src/server/db/uploads.js
```diff
@@ -41,11 +41,10 @@ function getUpload (id, trns = knex) {
 }
 
 function validForReportingPeriod (period_id, trns = knex) {
```
ajhyndman, line 43: What do you think about renaming this to `exportedForReportingPeriod`?  I was a little confused at first because I thought this function returned "all" valid uploads for the reporting period.

ajhyndman (approved):
Looks good to me

src/server/db/uploads.js
igor47: this would be a merge conflict with a PR you have open (auditing) so i'll do it as a follow-up.

PR #313, opened 2022-06-10
Support Subrecipient template in treasury export

ajhyndman:


src/server/db/arpa-subrecipients.js
```diff
@@ -65,10 +65,15 @@ async function listRecipients (trns = knex) {
   return baseQuery(trns)
 }
 
+async function listRecipientsForReportingPeriod (periodId, trns = knex) {
+  return baseQuery(trns).where('reporting_period_id', periodId)
+}
```
igor47, line 70: i wonder if this is sufficient? what if the subrecipient was modified in a subsequent period? maybe that's an audit feature? i'll open a separate issue to think about various kinds of marking we can do, or how else we handle sub-recipient changes.

igor47 (approved):
lgtm, thanks!

PR #314, opened 2022-06-11
Igor/standard form

igor47:
This is a cleanup PR which probably doesn't need to exist. We tackle two issues:

1. `RecordForm` is kind of a weird abstraction, with lots of configuration that's not being used
2. Our create/update flow for users and agencies was kinda odd, with updates attempting to happen in several places (on the backend, but also in the store) and functionality spread between several components and the store

Here, we switch to using a new abstraction -- `StandardForm` -- for users, agencies, and subrecipients. Additionally, combine create and update code paths for these, both on the front-end (via front-end routes) and the back-end (by using http `POST` for both). 

Unlike `RecordForm`, which requires passing update methods as props, `StandardForm` emits events for reset and save, which allows the parent component to handle update functionality.

We refactor some code out of the store, further reducing it in size. This PR definitely conflicts with usdigitalresponse/arpa-reporter#308 -- sorry about that!

We cannot yet remove `RecordForm` because it's still in use in the `ReportingPeriod` component. Refactoring `ReportingPeriod` warrants it's own PR, because the front-end state for this is a total mess, with lots of getters, difficult to follow initialization flow, and duplicate buts of state like `allReportingPeriods` and `reportingPeriods`. This would certainly require touching a bunch of code, including in the store and in dependent components, and so would further affect usdigitalresponse/arpa-reporter#308 

Recommend implementing usdigitalresponse/usdr-gost#396 on top of this PR, maybe by adding props to `StandardForm` like `enableDelete` (maybe also `enableCancel`?) and then handling the appropriate events.

### DIFF
--- a/src/components/StandardForm.vue
+++ b/src/components/StandardForm.vue
@@ -1,0 +1,111 @@
+<template>
+  <div>
+    <div class="form-group row" v-for="col in cols" :key="col.field">
+      <label :for="col.field" class="col-sm-2 col-form-label">{{ col.label }}</label>
+
+      <div class="col-sm-10" >
+        <select
+          v-if="col.selectItems && col.selectItems.length > 0"
+          class="form-control"
+          :id="col.field"
+          v-model="record[col.field]"
+          :readonly="col.readonly"
+          :class="classesForField(col.field)"
+          :aria-describedby="`feedback-${col.field}`"
+          >
+          <option v-for="opt in col.selectItems" :key="opt.value" :value="opt.value">
+            {{ opt.label }}
+          </option>
+        </select>
+
+        <input
+          v-else
+          type="text"
+          class="form-control"
+          :id="col.field"
+          v-model="record[col.field]"
+          :readonly="col.readonly"
+          :class="classesForField(col.field)"
+          :aria-describedby="`feedback-${col.field}`"
+          />
+
+        <div :id="`feedback-${col.field}`" class="invalid-feedback">
+          <span v-for="error in errors[col.field]" :key="error">
+            {{ error }}
+          </span>
+        </div>
+      </div>
+    </div>
+
+    <div class="form-group row">
+      <div class="col-sm-2"></div>
+      <div class="col-sm-10">
+        <button class="btn btn-primary" v-on:click="validateAndSave()" :disabled="disabled">Save</button>
+        <button class="btn btn-secondary ml-2" v-on:click="$emit('reset')" :disabled="disabled">Reset</button>
+      </div>
+    </div>
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'StandardForm',
+  props: {
+    initialRecord: Object,
+    cols: Array,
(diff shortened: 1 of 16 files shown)


src/server/routes/configuration.js
```diff
@@ -3,14 +3,12 @@
...
-const { user: getUser, users: getUsers, roles: getRoles } = require('../db/users')
+const { roles: getRoles } = require('../db/users')
 
 router.get('/', requireUser, async function (req, res) {
-  const user = await getUser(req.signedCookies.userId)
-  const users = user.role === 'admin' ? await getUsers() : [user]
   const roles = await getRoles()
 
-  res.json({ configuration: { users, roles } })
+  res.json({ configuration: { roles } })
```
igor47, line 11: we can almost kill this `configuration` route -- we should just keep track of roles in the store directly, vs as the only remaining child of the configuration object.

src/components/StandardForm.vue
```diff
@@ -0,0 +1,108 @@
...
+        // check required fields
+        if (col.required && (val === undefined || val === null || val === '')) {
+          this.errors[field].push(`${col.label} is a required field`)
+        }
+      })
+
+      this.wasValidated = true
+      const hasErrors = Object.values(this.errors).some(errList => errList.length > 0)
+      if (!hasErrors) {
+        this.$emit('save', this.record)
```
ajhyndman, line 103: Could you explain the advantage of events over callbacks is, as you see it?

It seems very much the same thing to me, only with an extra level of asynchrony introduced.

src/components/StandardForm.vue
```diff
@@ -0,0 +1,108 @@
...
+          :aria-describedby="`feedback-${col.field}`"
+          >
+          <option v-for="opt in col.selectItems" :key="opt.value" :value="opt.value">
+            {{ opt.label }}
+          </option>
+        </select>
+
+        <input
+          v-else
+          type="text"
```
ajhyndman, line 23: I don't feel that strongly, but I'd prefer to add a `type` attribute to each col.

Especially, if we ever add support for more than select and text inputs, I think it would help the code read more clearly.  Then we can switch explicitly on that type value when rendering inputs.

ajhyndman (approved):
I quite like this abstraction!  Seems to read pretty well, and it will definitely save some lines of code after #316.

I just left a couple optional suggestions.

src/components/StandardForm.vue
igor47: i think it makes it easier, in a parent component, to see when functions are being called. it makes sense to me that the parent component is responsible for the flow of data, and the form is just responsible for layout, ui, and emitting events

src/components/StandardForm.vue
igor47: i don't have a good source of the type data. in the follow-up pr, i add an `isDate: true` which is how `RecordForm` did it. so far, we have strings, dates, and selects.

Issue #315, opened 2022-06-11
kill configuration routes and store


igor47:
we should just keep track of roles in the store directly, vs as the only remaining child of the configuration object.

_Originally posted by @igor47 in https://github.com/usdigitalresponse/arpa-reporter/pull/314#discussion_r894998019_

igor47:
fixed in #316

PR #316, opened 2022-06-11
Refactor period management in the UI

### LINKED ISSUE #234: re-factor the store initialization logic
this logic is still buggy. in particular, we only seem to set `viewPeriodId` in `loadApplicationSettings` and it ends up being `null` a lot of the time, which kinda still works because it should be set to the current period, and the backend generally also uses the current period if the specified period is not set, but this is not ideal. the entire flow of how the app loads it's initial state into the store should be re-worked.

_Originally posted by @igor47 in https://github.com/usdigitalresponse/arpa-reporter/pull/233#discussion_r869519851_

igor47:
Follow-up to #314 that switches the store of reporting periods to follow the conventions used for users, subrecipients, and agencies. This includes switching to `StandardForm`, which allows us to delete the `RecordForm` components.

This PR fixes #234 -- the logic is now consistent, and we `await` all of the app's necessary state before rendering for a more consistent view.

As of this PR, we no longer duplicate update logic on the backend and the frontend. Instead, all backend updates force a reload of the state of the world into the frontend, to eliminate state synchronization bugs.


igor47:
fixes #315

src/store/index.js
```diff
@@ -162,87 +149,45 @@ export default new Vuex.Store({
...
-            }
-          })
-      }
-      doFetch('configuration')
-      doFetch('reporting_periods')
+      // to ensure consistent application state, lets block rendering until these complete
+      await dispatch('updateApplicationSettings')
+      await dispatch('updateReportingPeriods')
+      await dispatch('updateAgencies')
+      await dispatch('updateUsersRoles')
```
ajhyndman, line 159: Ooh, I think we can parallelize these and still block render.

```suggestion
      await Promise.all([
        dispatch('updateApplicationSettings'),
        dispatch('updateReportingPeriods'),
        dispatch('updateAgencies'),
        dispatch('updateUsersRoles')
      ])
```

src/store/index.js
```diff
@@ -261,6 +206,36 @@ export default new Vuex.Store({
...
+    async updateUsersRoles ({ commit, state }) {
+      const result = await getJson('/api/users')
+      if (result.error) {
+        commit('addAlert', { text: `updateUsersRoles Error: ${result.error} (${result.text})`, level: 'err' })
+      } else {
+        commit('setRoles', result.roles)
+        commit('setUsers', result.users)
+      }
+    },
+    async updateApplicationSettings ({ commit }) {
```
ajhyndman, line 230: I do really like this store change.  A lot easier to reason about, you're right.  Great work, thanks!

ajhyndman (approved):
I did confirm that this fixes #326!

Just left one minor suggestion.

PR #317, opened 2022-06-13
Generate an audit report!

ajhyndman:
So far this only adds a single sheet to the audit report.  That sheet lists aggregated obligations and expenditures per upload.

OZZIE'S PROPOSAL

Screen Shot 2022-06-12 at 5 08 59 pm

ACTUAL OUTPUT

(Note that the data is transposed compared to Ozzie's handwritten sheet.)

Screen Shot 2022-06-12 at 5 07 40 pm

Addresses usdigitalresponse/usdr-gost#397


src/server/lib/audit-report.js
```diff
@@ -1,876 +1,118 @@
-/* eslint camelcase: 0 */
-
-/*
---------------------------------------------------------------------------------
--                           lib/audit-report.js
---------------------------------------------------------------------------------
-
-*/
-const _ = require('lodash')
+const { format } = require('date-fns')
```
igor47, line 1: whoa has `date-fns` been in our bundle this whole time? maybe we should just stick to moment, it's used in many more places

src/server/lib/audit-report.js
```diff
@@ -1,876 +1,118 @@
...
 const XLSX = require('xlsx')
+
+const { get, getAll } = require('../db/reporting-periods')
```
igor47, line 4: maybe alias them? `get` is kinda hard to grep for, compared to, like, `getReportingPeriod`

igor47 (approved):
cool, lgtm, i left a few minor nits. i'm not sure about the `validForReportingPeriod` subquery -- that's the main thing that needs some thought.

src/server/lib/audit-report.js
ajhyndman: I'd honestly be happier with just exporting the verbose name.  How does that sound to you?

src/server/lib/audit-report.js
ajhyndman: You're definitely right.  I just cloned this from the previous version of the code.  I'll split off a follow-up!

src/server/db/uploads.js
```diff
@@ -46,6 +46,7 @@ function validForReportingPeriod (period_id, trns = knex) {
   ))
     .select('uploads.*')
     .from('uploads')
+    .where('uploads.reporting_period_id', period_id)
```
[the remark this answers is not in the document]
ajhyndman: I did test to verify that this filters correctly.  Is that what you mean?

src/server/lib/audit-report.js
ajhyndman: I had to rebase, so I snuck this in after all.  👍

PR #331, opened 2022-06-16
Igor/consolidate env

igor47:
As we bring up new environments for additional customers, the question arises: what environment config is necessary? In this PR, we make it easy to answer the question by centralizing all configuration into the `environment.js` dictionary.


ajhyndman (approved):
💯

PR #332, opened 2022-06-16
Refactor reporting period management

### LINKED ISSUE #309: Uploading the same UST template across reporting periods results in an error.
The error is "The file [filename] already exists."

igor47:
We do a few things here:

* Remove the oodles of unused columns on the `reporting_periods` table
* Make the `certified_by` a FK reference to users
* Change the way custom period templates are stored (which fixes #309 )
* Moves remaining single-purpose API interactions into relevant components (for uploading a template and certifying a reporting period)
* fix faulty logic around certification, which should fix the comment we heard this afternoon re "I always have to ask Larry to certify the period for me"

### DIFF
--- a/migrations/20220616180456_overhaul_reporting_periods.js
+++ b/migrations/20220616180456_overhaul_reporting_periods.js
@@ -1,0 +1,48 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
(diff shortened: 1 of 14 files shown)


ajhyndman (approved):
Thanks for confirming these extra columns could be cleaned up in today's meeting!

I had a similar thought when I was reviewing some of the multitenant changes, but didn't follow through.

> - fix faulty logic around certification, which should fix the comment we heard this afternoon re "I always have to ask Larry to certify the period for me"

Nice catch!

PR #333, opened 2022-06-17
Add links to uploads to Audit Report

### LINKED ISSUE #330: Audit Report: Publish a link to upload to a new column
Put a link to the upload details page in arpa-reporter directly in the audit report!

ajhyndman:
Fixes #330

### DIFF
--- a/src/server/environment.js
+++ b/src/server/environment.js
@@ -27,7 +27,6 @@ const IS_DEV = NODE_ENV === 'development'
 const LOGIN_EXPIRY_MINUTES = parseInt(process.env.LOGIN_EXPIRY_MINUTES) || 30
 
 const PORT = parseInt(process.env.PORT) || 3000
-const API_DOMAIN = process.env.API_DOMAIN
 const WEBSITE_DOMAIN = process.env.WEBSITE_DOMAIN
 
 const NOTIFICATIONS_EMAIL = process.env.NOTIFICATIONS_EMAIL
@@ -53,7 +52,6 @@ module.exports = {
   IS_DEV,
   LOGIN_EXPIRY_MINUTES,
   PORT,
-  API_DOMAIN,
   WEBSITE_DOMAIN,
   NOTIFICATIONS_EMAIL,
   NOTIFICATIONS_EMAIL_PW,
--- a/src/server/environment.js
+++ b/src/server/environment.js
@@ -26,7 +26,9 @@ const IS_DEV = NODE_ENV === 'development'
 
 const LOGIN_EXPIRY_MINUTES = parseInt(process.env.LOGIN_EXPIRY_MINUTES) || 30
 
-const PORT = process.env.PORT || 3000
+const PORT = parseInt(process.env.PORT) || 3000
+const API_DOMAIN = process.env.API_DOMAIN
+const WEBSITE_DOMAIN = process.env.WEBSITE_DOMAIN
 
 const NOTIFICATIONS_EMAIL = process.env.NOTIFICATIONS_EMAIL
 const NOTIFICATIONS_EMAIL_PW = process.env.NOTIFICATIONS_EMAIL_PW
@@ -51,6 +53,8 @@ module.exports = {
   IS_DEV,
   LOGIN_EXPIRY_MINUTES,
   PORT,
+  API_DOMAIN,
+  WEBSITE_DOMAIN,
   NOTIFICATIONS_EMAIL,
   NOTIFICATIONS_EMAIL_PW,
   NODEMAILER_HOST,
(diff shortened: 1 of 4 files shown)


src/server/environment.js
```diff
@@ -21,7 +21,9 @@ const IS_DEV = NODE_ENV === 'development'
 
 const LOGIN_EXPIRY_MINUTES = parseInt(process.env.LOGIN_EXPIRY_MINUTES) || 30
 
-const PORT = process.env.PORT || 3000
+const PORT = parseInt(process.env.PORT) || 3000
+const API_DOMAIN = process.env.API_DOMAIN
```
igor47, line 25: we already have a process.env.BASE_URL. i didn't deal with it in https://github.com/usdigitalresponse/arpa-reporter/pull/331 because it's only used client-side (in src/router/index)

but actually, i am not seeing `BASE_URL` set anywhere, including in dev or in the production environments, so i wonder if it can be safely removed

igor47 (changes requested):
rather than setting a few more env vars, we can probably just pull the domain out of the `req` variable used to kick off the audit report gen process. we might not have the website url available in the API, but we do have it available on the client-side when the request is made, maybe it could be part of the request parameters?

ajhyndman:
> rather than setting a few more env vars, we can probably just pull the domain out of the `req` variable used to kick off the audit report gen process. we might not have the website url available in the API, but we do have it available on the client-side when the request is made, maybe it could be part of the request parameters?

Yeah, there are definitely a few approaches here.  I don't have a clear preference.

(a) As noted, we can't use the default properties on the express request, because the API host isn't necessarily the same as the website host.
(b) We can declare the website host in an environment variable, but this an extra point of configuration to remember.
(c) We can derive the host from window.location in the client and pass it to the server in a request parameter.  The issue here is that we have to trust the client, and something like a browser extension could hijack either window.location or issue requests with an alternative parameter.

🤷

igor47:
> (a) As noted, we can't use the default properties on the express request, because the API host isn't necessarily the same as the website host.
only in dev, in prod they're the same

> (c) We can derive the host from window.location in the client and pass it to the server in a request parameter. The issue here is that we have to trust the client, and something like a browser extension could hijack either window.location or issue requests with an alternative parameter.
i don't think browser extensions hijack window.location, that's a pretty core security feature of browsers. i would prefer this approach, but i can unblock this approach also if you think it's much better, esp. if you also clean up the existing `BASE_URL` config.

ajhyndman:
Okay, I learned a few more things

First, `BASE_URL` is actually an env variable auto-generated by Vue.

https://cli.vuejs.org/guide/mode-and-env.html#using-env-variables-in-client-side-code

Bizarrely, it doesn't seem like variables we declare in `.env` are actually available to the Vue HTML processory, so I don't think we can remove `BASE_URL`.  Note that the GOST repo also uses BASE_URL for the same purposes.  (I'm pretty certain this is part of the boilerplate generated when you bootstrap a new Vue app.

https://github.com/usdigitalresponse/usdr-gost/search?q=BASE_URL

Of course, since this env variable is provided by vue-cli-service, we can't access it from express — especially in production.  I can't see a way around keeping both values.  😧 

---

Second, I'm pretty sure you're correct: modern browsers do a good job of protecting window.location, so that particular concern probably isn't a real problem.

I admit I might be a little paranoid, but this is still a service that we are deploying into government agencies, and I just don't like the idea of passing around URLs in query parameters where they can be accessed or spoofed.  If an attacker did manage to get an arbitrary value into that domain, it could take the user literally anywhere.

igor47:
path forward: lets set an env var. looks like `api_host` is not used, just `website_host`. lets have a function that gets it out of the request `host` header, and then falls back to the env var if it's not set. we don't have to set it in prod -- the host header always matches for us.

ajhyndman:
@igor47 Sorry, I don't think I quite follow you actually.

Are you suggesting that we read the default [`Host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/host) header from the request and only falling back to the env variable if that isn't present?  I think that header should always be present, it's just incorrect in our dev configuration.

image

image

igor47:
i suggest we try the env var and fall back to the `Host` header if that's not present. we'll only have to configure the `Host` header in dev.

ajhyndman:
Ohh, reversing the direction of the fallback.  Sure, that makes more sense.  👍

src/server/lib/audit-report.js
```diff
@@ -19,17 +20,23 @@ const COLUMN = {
   E_CPE: 'Current Period Expenditures (Aggregate Awards <50k)'
 }
 
-async function generate () {
+function getUploadLink (domain, id, filename) {
+  return { f: `=HYPERLINK("${domain}/uploads/${id}","${filename}")` }
+}
+
+async function generate (requestHost) {
```
igor47, line 27: just need to change where you're invoking `generate` to pass `requestHost`

igor47 (changes requested):

igor47 (approved):
thanks, lgtm.

PR #334, opened 2022-06-17
fix: error on logged-out view

igor47:
`viewPeriod` is not available if we're not logged in


igor47:
i'm merging this without waiting for approval to fix the existing sites.

PR #340, opened 2022-06-18
Crosswalk between projects and uploads in audit report

ajhyndman:
We probably want to do more with this tab eventually, but mapping between projects and uploads has been highly requested, so we solve that now.

Fixes #340

### DIFF
--- a/src/server/lib/audit-report.js
+++ b/src/server/lib/audit-report.js
@@ -3,7 +3,7 @@ const XLSX = require('xlsx')
 
 const { get, getAll } = require('../db/reporting-periods')
 const { getCurrentReportingPeriodID } = require('../db/settings')
-const { recordsForUpload } = require('../services/records')
+const { recordsForReportingPeriod } = require('../services/records')
 const { log } = require('../lib/log')
 const { usedForTreasuryExport } = require('../db/uploads')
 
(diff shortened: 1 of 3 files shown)


src/server/services/persist-upload.js
```diff
@@ -50,7 +50,7 @@ async function persistUpload ({ filename, user, buffer }) {
   return upload
 }
 
-function bufferForUpload (upload) {
+async function bufferForUpload (upload) {
```
igor47, line 53: just curious, why is this necessary?

src/server/lib/audit-report.js
```diff
@@ -42,77 +50,120 @@ async function createObligationSheet (periodId) {
...
+    switch (record.type) {
+      case 'ec1':
+      case 'ec2':
+      case 'ec3':
+      case 'ec4':
+      case 'ec5':
+      case 'ec7':
+        rows.push({
+          'Project ID': record.content.Project_Identification_Number__c,
+          'Upload': record.upload.filename // TODO: link to upload in arpa reporter
```
igor47, line 155: i'm not sure how useful the filename will be, seems like maybe including at least an upload id column would be convenient, a link would be even better.

src/server/services/records.js
```diff
@@ -36,8 +37,8 @@ function extractRecords (buffer) {
   const subcategory = cover['Detailed Expenditure Category']
 
   const records = [
-    { type: 'certification', content: certification },
-    { type: 'cover', content: cover }
+    { type: 'certification', upload, content: certification },
+    { type: 'cover', upload, content: cover }
```
igor47, line 41: hmm, do these types of record not have a `subcategory` field?

igor47 (approved):

src/server/services/persist-upload.js
ajhyndman: I don't think it is, actually.

I was fiddling with things because I noticed that bufferForUpload was running in series instead of in parallel when called from `recordsForReportingPeriod`.  I'm still not 100% sure why, but inlining `extractRecords` to `recordsForUpload` seemed to do the trick.

I guess I left it in because otherwise there's nothing indicating that this function is async without reading more code.

src/server/lib/audit-report.js
ajhyndman: Yeah, I'm imagining linking the filename to the upload in the tool.  👍 

I'll fix that as soon as we decide on a solution for links.

src/server/services/records.js
ajhyndman: Hmm, that's true.  I'm not sure we actually use these records for anything other than extracting the subcategory at the moment.  Maybe we should just omit them?

PR #343, opened 2022-06-20
Validation warnings

### LINKED ISSUE #327: Subrecipients: Don't block uploads when subrecipient details don't match
Let's aim for parity with CARES reporter:

- First upload that specifies a Subrecipient is captured
- Duplicate Subrecipients in subsequent reports are ignored
- (Optional) Report simplified "warnings" during the upload process

igor47:
fixes #327 

we display mismatch errors as warnings in the UI. we still treat missing fields as fatal errors.

### DIFF
--- a/src/server/lib/validation-error.js
+++ b/src/server/lib/validation-error.js
@@ -1,6 +1,6 @@
 
 class ValidationError extends Error {
-  constructor (message, { severity = 1, tab = null, row = null, col = null } = {}) {
+  constructor (message, { severity = 'err', tab = null, row = null, col = null } = {}) {
     super(message)
     this.severity = severity
     this.tab = tab
--- a/src/server/lib/validation-error.js
+++ b/src/server/lib/validation-error.js
@@ -1,4 +1,7 @@
 
+/**
+ * severity: "err" | "warn"
+ */
 class ValidationError extends Error {
   constructor (message, { severity = 'err', tab = null, row = null, col = null } = {}) {
     super(message)
--- a/src/server/services/validate-upload.js
+++ b/src/server/services/validate-upload.js
@@ -104,7 +104,7 @@ async function validateSubrecipientRecord ({ upload, recipient, rules, trns }) {
   } else {
     errors.push(new ValidationError(
       'At least one of UEI or TIN must be set, but both are missing',
-      { col: 'C, D' }
+      { col: 'C, D', severity: 'err' }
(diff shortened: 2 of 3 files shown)


src/server/lib/validation-error.js
```diff
@@ -1,6 +1,6 @@
 
 class ValidationError extends Error {
-  constructor (message, { severity = 1, tab = null, row = null, col = null } = {}) {
+  constructor (message, { severity = 'err', tab = null, row = null, col = null } = {}) {
```
ajhyndman, line 3: Nit: I like using descriptive strings over numbers, but I also have a pet peeve for unnecessary abbreviations.  Could we just call it `'error'`?

src/server/services/validate-upload.js
```diff
@@ -31,9 +31,10 @@ async function validateAgencyId ({ upload, records, trns }) {
     )
   }
 
-  // set agency id on the upload, for disambiguation
+  // always set agency id if possible; we omit passing the transaction in this
+  // case, so that the agency id gets set even if the upload fails to validate
   if (matchingAgency.id !== upload.agency_id) {
-    await setAgencyId(upload.id, matchingAgency.id, trns)
+    await setAgencyId(upload.id, matchingAgency.id)
```
ajhyndman, line 37: Just double-checking.  If the validation errors immediately above are triggered this should not run, correct?

src/server/lib/validation-error.js
```diff
@@ -1,6 +1,6 @@
 
 class ValidationError extends Error {
```
ajhyndman, line 2:
```suggestion
/**
 * severity: "err" | "warn"
 */
class ValidationError extends Error {
```

Could you add a short comment or something expressing which severities are expected?

ajhyndman (approved):
A couple minor requests, but looks good!

src/server/lib/validation-error.js
igor47: i'm mirroring what i did in https://github.com/usdigitalresponse/arpa-reporter/blob/3a3851f2f31b5580f1b86771424d6ecd111c61c6/src/components/AlertBox.vue#L20

src/server/services/validate-upload.js
igor47: yes, those immediately `return` with the error. we only get here if there was no error.

PR #344, opened 2022-06-20
Declare infra with render.com blueprint file

ajhyndman:
Use Render's blueprint syntax to declare the servies required to launch a new ARPA instance.

https://render.com/docs/blueprint-spec

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -14,6 +14,8 @@ services:
       name: DataDir
       mountPath: /var/data
       sizeGB: 10
+    # Environment variables. Please keep declaration in this file in sync with:
+    # /src/server/environment.js
     envVars:
       - key: AWS_ACCESS_KEY_ID
         sync: false # placeholder for a value to be added in the dashboard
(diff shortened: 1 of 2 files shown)


render.yaml
```diff
@@ -0,0 +1,41 @@
...
+    name: arpa-reporter-web
+    env: node
+    region: ohio
+    branch: main
+    buildCommand: yarn install && yarn build
+    startCommand: yarn run prod
+    healthCheckPath: /api/health
+    disk:
+      name: Uploads
+      mountPath: /var/data/uploads
```
igor47, line 15:
```suggestion
      name: DataDir
      mountPath: /var/data
```
i know that's not what we currently have, but i wonder if the entire data dir, not just the uploads sub-path, should be the volume...

igor47 (approved):
wow, this ends up being very compact. fewer env vars than i expected, too!

one minor additional suggestion: add comments both here and in `environment.js` linking future people to both files to help keep the env in-sync.

render.yaml
ajhyndman: Yeah, I like this idea. 👍  It seems more future proof.

PR #345, opened 2022-06-20
Remove dead fields from reporting periods seed

ajhyndman:


igor47 (approved):
oops, i missed these in #332 -- thanks!

PR #348, opened 2022-06-21
Igor/validation rules

igor47:
Initial pass at validation based on rule description in the input template.

We generate the rules as a JSON file by parsing the header of the input template, and save it in the repo. The rules can be re-generated if the spreadsheet is updated, by running `yarn generateRules`.

We fold most of subrecipient validation, which is duplicate, into the general-purpose rule validation code.

We currently validate the following:
* required fields are present
* max length is not exceeded
* a picklist value comes from the relevant pick list


igor47:
I implemented parsing the `Logic` sheet and set an `ecCodes` attribute on each rule, indicating the upload code to which the rule applies. this got rid of many errors, leaving what looks like relevant ones. i propose we ship this as-is -- probably, lots of sheets will fail to validate, and then we can figure out what to do about the remaining validation errors.

scripts/generateRules.mjs
```diff
@@ -0,0 +1,219 @@
+#!/usr/bin/env node
```
ajhyndman, line 1: I'm happy to land and then iteratie on this, but I would love to run this as a build step instead of committing the generated JSON.  Any objections?

package.json
```diff
@@ -22,7 +22,8 @@
     "test:server-aws-ses": "./tests/server-nodemailer/mocha_wrapper.sh",
     "lint": "vue-cli-service lint",
     "prod": "NODE_ENV=production node src/server",
-    "initRender": "./scripts/initRender.mjs"
+    "initRender": "./scripts/initRender.mjs",
+    "generateRules": "./scripts/generateRules.mjs"
```
ajhyndman, line 26: Thanks for adding an alias to package.json!

scripts/generateRules.mjs
```diff
@@ -0,0 +1,219 @@
...
+import { readFile, writeFile } from 'fs/promises'
+import path from 'path'
+
+import 'dotenv/config'
+import chalk from 'chalk'
+import XLSX from 'xlsx'
+import lodash from 'lodash'
+
+import { DATA_SHEET_TYPES } from '../src/server/services/records.js'
+import { SERVER_DATA_DIR, EMPTY_TEMPLATE_NAME, SRC_DIR } from '../src/server/environment.js'
```
ajhyndman, line 12: I guess we don't have to solve this now, but have you thought about how we might support custom templates per reporting period?

src/server/services/validate-upload.js
```diff
@@ -128,60 +122,118 @@ async function validateSubrecipientRecord ({ upload, recipient, rules, trns }) {
...
 }
 
-async function validateSubrecipients ({ upload, records, rules, trns }) {
+async function validateRules ({ upload, records, rules, trns }) {
   const errors = []
 
-  // validate each, and save the errors
-  const recipients = records.filter(rec => rec.type === 'subrecipient').map(r => r.content)
+  // go through every rule type we have
+  for (const [type, typeRules] of Object.entries(rules)) {
```
ajhyndman, line 199: In my head it feels more natural to iterate through records first and then find all rules that apply for that record.

I guess it's functionally identical though?  🤷  Feel free to ignore this.

scripts/generateRules.mjs
```diff
@@ -0,0 +1,219 @@
...
+    const type = DATA_SHEET_TYPES[sheetName]
+    const sheet = workbook.Sheets[sheetName]
+
+    // entire sheet
+    const sheetRange = XLSX.utils.decode_range(sheet['!ref'])
+
+    // range A1:13
+    const headerRange = merge({}, sheetRange, {
+      s: { c: 0, r: 0 },
+      e: { r: 12 }
```
ajhyndman, line 78: Nit: I guess an incremental step to avoiding magic numbers would be to declare the number of header rows as a constant somewhere, and share it across callsites.

src/server/services/validation-rules.js
```diff
@@ -1,105 +1,35 @@
...
-const { bufferForUpload } = require('./persist-upload')
-const { DATA_SHEET_TYPES } = require('./records')
-const { templateForPeriod } = require('./get-template')
+const srcRules = require('../lib/templateRules.json')
+const srcDropdowns = require('../lib/templateDropdowns.json')
 
-const COLNAMES = makeColNames()
-const STATES = ['AL', 'AK', 'AS', 'AZ', 'AR', 'CA', 'CO', 'CT', 'DE', 'DC', 'FM', 'FL', 'GA', 'GU', 'HI', 'ID', 'IL', 'IN', 'IA', 'KS', 'KY', 'LA', 'ME', 'MH', 'MD', 'MA', 'MI', 'MN', 'MS', 'MO', 'MT', 'NE', 'NV', 'NH', 'NJ', 'NM', 'NY', 'NC', 'ND', 'MP', 'OH', 'OK', 'OR', 'PW', 'PA', 'PR', 'RI', 'SC', 'SD', 'TN', 'TX', 'UT', 'VT', 'VA', 'VI', 'WA', 'WV', 'WI', 'WY']
+function generateRules () {
+  const rules = JSON.parse(JSON.stringify(srcRules))
```
ajhyndman, line 6: What happens if you omit the JSON.stringify/parse calls here?  I think it should be possible to avoid this extra serialize/deserialize pass.

ajhyndman (approved):
I'm happy to unblock!  Lets get this onto staging so that we can do some testing.  👍 

I left a few comments, but nothing needs to be blocking.

scripts/generateRules.mjs
igor47: i really hope we don't have to support custom sets of rules per reporting period, though the cares reporter code definitely had some provisions for doing that

src/server/services/validate-upload.js
igor47: it's not, because of [here](https://github.com/usdigitalresponse/arpa-reporter/blob/d80772bf5fb799f3c5d383110a20096b37c5094e/src/server/services/validate-upload.js#L232). we assume that if we filter records by type, the order in which they appear will be the order they appear in their original sheet. if we want to go through records first, and then rules, then we should save the row in the input spreadsheet in the record entry

PR #366, opened 2022-06-28
add node_version file

### LINKED ISSUE #64: set/maybe upgrade the nodejs version

igor47:
render actually uses this file to specify the node version in the
runtime environment, and it will help new developers have the correct
version set up locally

this is an alternative to #271 and fixes #64


ajhyndman:
Do you have a reason for preferring a `.node_version` file to #271 ?

Render respects both, and `yarn` also enforces compliance with the "engines" key, which I think is a nice win.  We can list an exact version in either place, if desired.

igor47:
yes, i don't think the yarn engines setting actually puts the right version of node into my path, which a `.node_version` file does. i have several projects on my laptop using different versions of node, and i've been keeping this file in my repo (out of version control) for the whole time anyway.

ajhyndman:
Oh, cool.  What tool are you using that respects .node_version files like that?  Does nvm do that now?

I think last time I tried to do something like this with a project we found that you still had to call `nvm use` to update your path.

igor47:
> Oh, cool. What tool are you using that respects .node_version files like that?

i use `asdf` and `direnv`. here's the best guide i was able to find to replicate my setup:
https://blog.sighup.io/manage-tools-with-ease-direnv-asdf/

i highly recommend it. for arpa-reporter, i have a (globally git-ignored) `.envrc` file that contains:
```
use asdf
export RENDER_API_KEY="rnd_<redacted>"
```

this way, not only do i switch to the correct version of node when i CD into that project directory, but i also have a render API key available (which i used for the migration scripts). the API key is unloaded when i CD out of that directory.

ajhyndman (approved):
Okay cool.  I really wish yarn would enforce compliance with .node-version files too.  That way nobody could accidentally miss our setup directions.

I guess standards are hard though.  🤷

mattbroussard:
- `.node_version`: Render, Yarn, `asdf`, `nodenv`
- `engines`: Render, Yarn, maybe `nvm` (seem like there is some [debate](https://github.com/nvm-sh/nvm/issues/651) about whether `nvm` should support it, not sure what the current state is)
- `.nvmrc`: `nvm`

Since `nvm` is the odd one out, maybe we should drop support for it and explicitly specify devs should use `asdf` or `Nodenv`. Right now ARPA readme says `nvm` or `asdf` and GOST readme says `nvm` or `nodenv` (I am personally using `nodenv` successfully in both repos). Or we could just keep multiple config files around and be sure to update them all when updating Node.

ajhyndman:
Yeah, I don't feel too strongly here.  I think we can just pick one.

My personal preference would be "engines", because yarn will _enforce_ compliance with that.  The other config files can be autodetected and applied if you have the right environment, but with "engines" nobody will be able to yarn install without a valid version of node installed.

mattbroussard:
Ah, fair. So I we should have `engines` _and_ `.node-version` at least, and optionally `.nvmrc` as well. GOST already has one so probably keeping that to avoid disrupting anyone who uses `nvm` will be best.

PR #372, opened 2022-06-28
Pass tenant ID everywhere required in audit report

ajhyndman:


igor47 (approved):
i wonder where else we missed some spots. should we try to look for all call to all functions that take a tenantId and make sure they have one?

PR #373, opened 2022-06-29
Encode some basic DB workflows with scripts

ajhyndman:
Hopefully this helps make some of our common DB workflows more discoverable and repeatable!

- Remove broken `refreshdb` script
- Rewrite `cleandb` script to respect `.env`
- List common db commands as scripts in `package.json`


scripts/cleandb.sh
```diff
@@ -0,0 +1,18 @@
...
+# Get DB name from connection URI
+dbname=${POSTGRES_URL#*//*/}
+# Remove DB name from connection URI
+postgres_url=${POSTGRES_URL/$dbname}
+
+echo $dbname
+
+psql $postgres_url -c "DROP SCHEMA public CASCADE"
+psql $postgres_url -c "CREATE SCHEMA public"
+psql $postgres_url -c "DROP DATABASE IF EXISTS $dbname"
```
igor47, line 17: why do we need to drop the schema if we can just drop the database? also, why do we need to drop the database *after* dropping the schema? i'm not sure how the `IF EXISTS` could ever be true here.

igor47 (approved):
lgtm, thanks, left a minor comment.

scripts/cleandb.sh
ajhyndman: Oh, thanks!  This was some lazy copy pasta on my part from `tests/server/reset-db.sh`.  It's been a while since I did a lot of SQL work.

PR #374, opened 2022-06-29
Document deployment process with RENDER.md

ajhyndman:
We should remove this file when merging ARPA and GOST repositories.


igor47 (approved):
lgtm; should we leave a reference to this document in the `README` ?

PR #383, opened 2022-06-30
Give human-readable validation when missing EC code

trumanc:
Fixes the exception error message that is returned when the validation code tries to call `.split()` on an empty ec-code cell value. It checks for the empty value and returns a human-readable validation error before the exception can be thrown.

### DIFF
--- a/src/server/services/validate-upload.js
+++ b/src/server/services/validate-upload.js
@@ -19,7 +19,7 @@ async function validateAgencyId ({ upload, records, trns }) {
 
   // must be set
   if (!agencyCode) {
-    return new ValidationError('Agency code must be set', { tab: 'cover', row: 1, col: 0 })
+    return new ValidationError('Agency code must be set', { tab: 'cover', row: 1, col: 'A' })
   }
 
   // must exist in the db
@@ -27,7 +27,7 @@ async function validateAgencyId ({ upload, records, trns }) {
   if (!matchingAgency) {
     return new ValidationError(
       `Agency code ${agencyCode} does not match any known agency`,
-      { tab: 'cover', row: 2, col: 1 }
+      { tab: 'cover', row: 2, col: 'A' }
     )
   }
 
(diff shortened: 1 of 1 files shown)


trumanc:
Original error message looked like this:
Screen Shot 2022-06-28 at 10 20 51 AM


After the update, the same validation call now gives:
Screen Shot 2022-06-30 at 2 33 15 PM

src/server/services/validate-upload.js
```diff
@@ -43,6 +43,13 @@ async function validateEcCode ({ upload, records, trns }) {
   const coverSheet = records.find(doc => doc.type === 'cover').content
   const codeString = coverSheet['Detailed Expenditure Category']
 
+  if (codeString === undefined) {
```
igor47, line 46: is there a reason to specifically check for `undefined` ? would `if(!codeString)` work?

src/server/services/validate-upload.js
```diff
@@ -43,6 +43,13 @@ async function validateEcCode ({ upload, records, trns }) {
...
+    return new ValidationError(
+      'Record is missing EC code',
```
igor47, line 48: not sure if `"Record"` is the right term here. maybe `"Upload"` ? also, maybe something like `"EC code must be set"` work match the missing-agency error well?

src/server/services/validate-upload.js
```diff
@@ -43,6 +43,13 @@ async function validateEcCode ({ upload, records, trns }) {
...
+      { tab: 'cover', row: 2, col: 4 }
```
igor47, line 49: possibly unrelated, our other errors generally refer to the column by it's letter, as that's how they show up in excel. these early validations on the cover sheet are now the odd ones out; in this case, i think it should be `col: 'D'`

igor47 (approved):
looks good! i left minor feedback, but this is already an improvement over the raw exception. feel free to merge when you're ready, and it'll get deployed to the staging environment.

src/server/services/validate-upload.js
trumanc: ah good call. The falsy evaluation is nicer because it will catch empty strings and nulls too. updating now

src/server/services/validate-upload.js
```diff
@@ -19,15 +19,15 @@ async function validateAgencyId ({ upload, records, trns }) {
 
   // must be set
   if (!agencyCode) {
-    return new ValidationError('Agency code must be set', { tab: 'cover', row: 1, col: 0 })
+    return new ValidationError('Agency code must be set', { tab: 'cover', row: 1, col: 'A' })
```
trumanc, line 22: This was inconsistent: This label assumed it was 0-indexed but everything else assumed A=1 and counted from there

trumanc:
New error messages, including the consistent language and col numbers changes to letters
Screen Shot 2022-06-30 at 3 38 53 PM
:

PR #387, opened 2022-07-01
fix: ensure validation always closes a transaction

igor47:
previously, we only closed it if the upload changed state, leaving db
connections hanging


igor47:
merging to unbreak staging

Issue #300, opened 2022-05-27
Ozzie: Can we make the ‘download’ icon a little more clear?


ajhyndman:
On both the Uploads and Upload screen, users have reported not being able to find the download button.

Maybe we can make this more visible, or consider experimenting with a different icon?

trumanc:
Fixed in #389

PR #389, opened 2022-07-04
Replace dl-icon with clearly labeled button

trumanc:
https://github.com/usdigitalresponse/arpa-reporter/issues/300 mentions that users have reported not being able to find the download button. This PR replaces the nondescript download icon with a clear labeled action button that is much harder to miss.

| old | new |
--- | ---|
| Screen Shot 2022-07-04 at 2 23 40 PM | Screen Shot 2022-07-04 at 2 24 08 PM | 
Screen Shot 2022-07-04 at 2 23 47 PM | Screen Shot 2022-07-04 at 2 24 01 PM |


src/components/DownloadFileButton.vue
```diff
@@ -1,18 +1,14 @@
 <template>
   <a :href="`/api/uploads/${upload.id}/download`">
-    <img
-      class="mr-4"
-      src="/assets/img/cloud-download.svg"
-      alt=""
-      width="20"
-      height="20"
-    />
+    <button class="btn btn-primary ml-2">
```
igor47, line 3: you don't need to embed a `button` inside an `a` anchor. instead, use button classes on the anchor itself. https://getbootstrap.com/docs/4.6/components/buttons/#button-tags

igor47 (approved):
lgtm, thanks, left a minor improvement suggestion

src/components/DownloadFileButton.vue
trumanc: updated to inline the button class on the anchor tag, and also removed the now unused svg file

src/components/DownloadFileButton.vue
```diff
@@ -0,0 +1,14 @@
+<template>
+  <a :href="`/api/uploads/${upload.id}/download`" class="btn btn-primary ml-2">
```
ajhyndman, line 2: You could try adding the class `btn-sm` as well.  Maybe that'll help prevent content from wrapping quite so easily?

ajhyndman (commented):
I think it might actually be sufficient to swap to a more recognizable icon, but this is probably an improvement.  👍 

Material Icons:

image

trumanc:
> I think it might actually be sufficient to swap to a more recognizable icon
Yeah I started with that idea but I figured it was better to make this button TOO visible instead of going with an icon that users might still miss. Better to fix this once instead of waiting for more user reports

Screen Shot 2022-07-05 at 11 29 28 AM

After making the button `btn-sm` the screen now looks like:

PR #395, opened 2022-07-05
Igor/validation fixes ozzie

igor47:
Includes 4 fixes from Ozzie's email this morning:
* enforce maxlength only on `String`-type fields
* check for missing values using an explicit list
* ignore case when validating pick list/multi-select options
* set the `tab` field of validation errors back to the sheet name

### DIFF
--- a/src/server/services/validate-upload.js
+++ b/src/server/services/validate-upload.js
@@ -229,8 +229,18 @@ async function validateRecord ({ upload, record, typeRules: rules, trns }) {
       continue
     }
 
+    // if the field is unset/missing/blank, is that okay?
+    if ([undefined, null, ''].includes(record[key])) {
+      // make sure required keys are present
+      if (rule.required === true) {
+        errors.push(new ValidationError(
+          `Value is required for ${key}`,
+          { col: rule.columnName, severity: 'err' }
+        ))
+      }
+
     // if there's something in the field, make sure it meets requirements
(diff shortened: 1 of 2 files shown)


src/server/services/validate-upload.js
```diff
@@ -225,29 +225,43 @@ async function validateRecord ({ upload, record, typeRules: rules, trns }) {
   // check all the rules
   for (const [key, rule] of Object.entries(rules)) {
     // if the rule only applies on different EC codes, skip it
-    if (rule.ecCodes && (!upload.ec_code || rule.ecCodes.indexOf(upload.ec_code) < 0)) {
+    if (rule.ecCodes && (!upload.ec_code || !rule.ecCodes.includes(upload.ec_code))) {
```
ajhyndman, line 228: 👍

ajhyndman (approved):
LGTM

src/server/services/validate-upload.js
```diff
@@ -225,29 +225,43 @@ async function validateRecord ({ upload, record, typeRules: rules, trns }) {
...
       continue
     }
 
+    // if the field is unset/missing/blank, is that okay?
```
trumanc, line 232: Might be useful to expand this comment to mention clearly that a numeric `0` is not an unset value, which is why we aren't using the normal falsy evaluation

trumanc (approved):

PR #397, opened 2022-07-05
Treasury exports v2022.06.21

ajhyndman:
- [x] Save new treasury templates to repository
- [x] Update references to template names in code
- [x] Update export logic to handle all new template mappings
- [x] Update export logic to handle all new columns


igor47 (approved):
cursory scroll and nothing jumps out at me; i'm not sure what a more detailed review would look like.

PR #399, opened 2022-07-06
feat: updated input template from joe as of today

### LINKED ISSUE #393: Update input template with version newer than July 1st
- [x] Joe to make changes to input template
- [ ] Update file in repository
- [ ] Regenerate validation rules

igor47:
updated based on latest guidance from treasury

closes #393


src/server/lib/templateDropdowns.json
```diff
@@ -126,31 +127,31 @@
...
-    "Sanitation disinfection and cleaning work",
-    "Social services work",
-    "Solid waste or hazardous materials management response and cleanup work",
-    "Transportation and warehousing",
-    "Vital services to Tribes",
-    "Work at hotel and commercial lodging facilities that are used for COVID-19 mitigation and containment",
-    "Work in a mortuary",
-    "Work in critical clinical research development and testing necessary for COVID-19 response",
-    "Work requiring physical interaction with patients",
+    "any work performed by an employee of a State local or Tribal government",
```
igor47, line 130: @jcomeau1 correct that these became lower-case again?

src/server/lib/templateDropdowns.json
```diff
@@ -218,14 +219,14 @@
     "Other (please specify)"
   ],
   "Subrecipient": [
-    "Subrecipient",
-    "Contractor",
-    "Beneficiary"
+    "-Subrecipient;",
```
igor47, line 222: @jcomeau1 what's up with these `-` and `;` at the front/end of these?

src/server/lib/templateRules.json
```diff
@@ -1255,6 +1255,32 @@
       "ecCodes": [
         "1.11"
       ]
+    },
+    "Cancellation_Reason__c": {
+      "key": "Cancellation_Reason__c",
+      "index": 47,
+      "required": "Conditional",
```
igor47, line 1262: might be interesting to implement this conditional require in the code. i always assumed we'd (in `validation-rules.js` method `generateRules`) would set `required` to a function, something like:

```javascript
rules.ec1.Cancellation_Reason__C.required = (record) => record.project_status_tbd === 'Cancelled'
```

src/server/lib/templateRules.json
```diff
@@ -3442,8 +3471,8 @@
         "2.1"
       ]
     },
-    "Number_Workers_Completing_Sectoral__c": {
-      "key": "Number_Workers_Completing_Sectoral__c",
+    "Number_Workers_Competing_Sectoral__c": {
```
igor47, line 3474: @jcomeau1 yup this is pretty maddening

src/server/lib/templateRules.json
```diff
@@ -5503,7 +5612,7 @@
       "key": "Is_project_designed_to_meet_100_mbps__c",
       "index": 33,
       "required": "Conditional",
-      "dataType": "TBD",
+      "dataType": "Pick List",
```
igor47, line 5615: this has become a `Pick List` but the `listVals` still read `TBD`, meaning that will be the only value permitted by render. we should source the pick list values now @jcomeau1

src/server/lib/templateRules.json
```diff
@@ -3800,7 +3878,7 @@
       "index": 16,
       "required": true,
       "dataType": "Numeric",
-      "maxLength": 6,
+      "maxLength": 10,
```
igor47, line 3881: you used to only be able to rehire 999,999 FTEs, but now I guess you can go up to 9,999,999,999 . that's larger than the population of the world, so we shouldn't need to increase this maxlength again.

unless, of course, they expect the commas. i'm not really sure what `Numeric` means, but i've been burned ever since the subrecipient `uei` was numeric -- length 10, and that meant a format like `\d{2}-\d{7}`

src/server/lib/templateRules.json
igor47: Maybe this is a yes/no ?

src/server/lib/templateDropdowns.json
jcomeau1: I have a note out to the partners to help us verify if the Guidance or the output template is correct in terms of capitalization.

src/server/lib/templateRules.json
jcomeau1: It would be good. There's a lot of the conditional logic already embedded within the input template as conditional formatting. While this is only a visual treatment within Excel, it might offer you a template to pass information to this kind of function.

src/server/lib/templateRules.json
jcomeau1: Yeah, it drives me nuts. But I bet they'll fix it the day before the period closes :)

src/server/lib/templateRules.json
jcomeau1: It is a Yes/No. I wonder if you have the ability to see the 'Data Validation' of a field? If so, it would give you the appropriate reference within the Dropdowns sheet.

Separately, I have replaced TBD with the values. I'll proved the revision once I hear back from the partners on the above questions.

src/server/lib/templateRules.json
jcomeau1: Treasury does NOT want commas. There was actually no max length listed, so I just made it consistent with other numeric fields. Call it future proofing.

src/server/lib/templateRules.json
jcomeau1: I added the metadata to the slightly revised verion I just sent. I would attach it, but unfortunately GitHub does not allow this. Check your email or Google Drive.

src/server/lib/templateDropdowns.json
ajhyndman: Yeah, we can definitely do normalization during output!

Any insight into why there are hyphens _and_ semicolons?

ajhyndman (approved):
If you want to land this for now, the changes look good to me.

jcomeau1:
I updated the metadata for EC5 cols AP and AQ and I also updated the metadata value list for EC4 col N. Same version number (v20220705), same location: https://drive.google.com/drive/folders/1HCRiHQUrgS2EIZrWmikngma8sKo5pSMS?usp=sharing

PR #400, opened 2022-07-06
Format Yes/No fields in treasury output

### LINKED ISSUE #394: Format Yes/No fields in exports

ajhyndman:
Define and use a new "capitalize" helper in our output generation.

Fixes #394

### DIFF
--- a/src/server/lib/format.js
+++ b/src/server/lib/format.js
@@ -24,7 +24,7 @@ const EXPENDITURE_CATEGORIES = {
  * @returns {string}
  */
 function capitalizeFirstLetter (value) {
-  if (value == null) return value
+  if (value == null || value === '') return value
   return `${value[0].toUpperCase()}${value.slice(1).toLowerCase()}`
 }
 


src/server/lib/format.js
```diff
@@ -13,32 +16,39 @@ const EXPENDITURE_CATEGORIES = {
...
-    return value
-  }
+/**
+ * Normalize casing of single word values.
+ * This is especially useful for "Yes"/"No" responses.
+ *
+ * @param {string} value
+ * @returns {string}
+ */
+function capitalize (value) {
```
trumanc, line 26: nit: Would it make sense to name this `capitalizeFirst` or something similar to disambiguate it from something like `toUpperCase()`?

src/server/lib/format.js
```diff
@@ -13,32 +16,39 @@ const EXPENDITURE_CATEGORIES = {
...
+  if (value == null) return value
```
trumanc, line 27: Probably need to check for the empty string here as well, or the `value[0].toUpperCase()` could throw an exception

src/server/lib/format.js
```diff
@@ -13,32 +16,39 @@ const EXPENDITURE_CATEGORIES = {
...
 
 function ec (value) {
   return EXPENDITURE_CATEGORIES[value]
 }
 
-function zip (value, optional = false) {
-  if (optional && value == null) {
-    return value
-  }
+function zip (value) {
```
trumanc, line 40: @ajhyndman  Did you mean to include all these optional/required changes in this PR that only mentions the Yes/No capitalization?

ajhyndman: Ah, yes.  Sorry, I included a small refactor to simplify things.

https://github.com/usdigitalresponse/arpa-reporter/pull/400/commits/231231de5ceb7036e5690ce1c8a7c5765d595f03

src/server/lib/format.js
ajhyndman: In JS, `''[0]` just returns undefined.

I'm not sure it's worth being too defensive in this code path.  If we encounter invalid input values, the right place to warn users is the validation stage.  Once we get here, we need to output something.

src/server/lib/format.js
ajhyndman: Oof, you're right.  This name is ambiguous.  I'll update!

src/server/lib/format.js
trumanc: the index won't throw, but calling toUpperCase on the undefined response will.

I'm not too worried about having particualarly strong validations here, but it seems like a weird edge case where a null value is fine but an empty string value crashes is nonintuitive and likely to cause confusion at some point.

trumanc (approved):
Up to you if you want to check for the empty string. Otherwise, LGTM

src/server/lib/format.js
ajhyndman: Oops, great point.  I guess I better fix that!

ajhyndman:
@igor47 I did a manual pass opening up each export template file and searching for "yes" in the text of the sheet.  To update, I'd end up doing the same thing.

Yes, theoretically, we could generate the formatting rules along with our other rules.  Form my point of view, someone has to manually define the rule to apply to each logical column anyway, and I think putting that mapping in the input template just moves it further away from the code.

The less frequently we have to update the input template for partners, the better I think the experience is.

PR #401, opened 2022-07-06
Upgrade vue (and related dependencies) to 2.7.2

ajhyndman:
Attempts to resolve errors from #391


trumanc (approved):
Looks like the original PR to bump to version 2.7.2 has been superseded a couple times since then, and the current version is 2.7.4 (https://github.com/usdigitalresponse/arpa-reporter/pull/412).

Do we want to bump this PR to that, or would you rather get this minor version bumped and then hope that the revision bum from 2.7.2 to 2.7.4 can be done more smoothly?

ajhyndman:
@trumanc I'm hoping it will be sufficient to land this PR and then let dependabot bump from 2.7.2 to 2.7.4.

I think going from 2.6.x to 2.7.x is the transition that dependabot isn't able to make automatically.

For the record, all I did here was flush the yarn.lock file and rebuild it so that transitive dependencies of the `vue` npm package would also be picked up.

PR #407, opened 2022-07-07
Format multi-select fields to normalize delimiters

ajhyndman:
- [x] Write multiselect helper
- [x] Apply helper to the two known fields
- [x] Write test cases for all formatting helpers


igor47 (approved):
lgtm; we definitely have two separate formatter systems now, maybe that's fine, maybe we could merge them at some point. :shrug:

Issue #419, opened 2022-07-12
[Partner submission] One of our agency users is getting the following error: 	Error	Record EC code 2.16 (Long-term Housing Security: Services for Unhoused Persons) does not match any known EC code


usdr-grants:
Link to submission is here:

https://ri-arpa-reporter-web.onrender.com/uploads/6c277d9a-06fd-48d8-b24c-579253abf321
https://www.jotform.com/uploads/grants.dev/212695501572052/5334437648218682894/prep1.PNG

jmartin-oh:
I am getting a similar error for 5.12
![image](https://user-images.githubusercontent.com/107556199/178516138-0ae26ccc-df87-4670-8bda-5358572bf85b.png)

jcomeau1:
Not sure what's happening with these codes. They do appear to be correct based on the list on the Logic sheet. Unfortunately, Ozzie's image didn't come through.

jmartin-oh:
For 5.12 there is a comma in the User Guide wording - I did try to add that manually just in case, but that didn't work either! There is not a comma in the Treasury's template, but thought I might give it try if we had something older in the code!

Jessica Martin, CIA, CGAP
Financial Manager
Ohio Grants Partnership

trumanc:
Is this EC code case sensitive?
I'm seeing here (https://github.com/usdigitalresponse/arpa-reporter/blob/staging/src/server/lib/arpa-ec-codes.js#L31) that we expect `Unhoused [p]ersons` to be lower case, but Ozzie's screenshot for code 2.16 is capitalized.

For code 5.12 in the first screenshot, the name is significantly different: https://github.com/usdigitalresponse/arpa-reporter/blob/staging/src/server/lib/arpa-ec-codes.js#L71

trumanc:
I can fix those. Should I check all the other EC codes for accuracy? What should I treat as the source-of-truth for what these names should be?

jmartin-oh:
Thank you!!! I think the bulk upload templates themselves, what is listed in the Sub_Category__C Column Notes as options. 

I just tried testing a bulk upload using what the User Guide indicates is the EC Name and I got errors.

trumanc:
I've confirmed I've found at least one other outdated name (2.25) so I'm going to carefully go through all templates

jcomeau1:
Sorry, I'm late to the game. All trailing spaces should be removed.

Truman, you should be able to rely on what's in the Input Template on the Logic sheet.

[...]

PR #422, opened 2022-07-12
Update EC code names for most recent templates

trumanc:
Got reports of 2 EC codes that weren't being recognized when partners tried to upload their files in https://github.com/usdigitalresponse/arpa-reporter/issues/419

I combed through all the bulk upload templates to find other cases of mismatches (did cmd-f searches for each string exactly as it appeared in the template) and found those two cases, plus 8 more mismatches.

~Note: Still waiting for an answer about code 2.34. The template includes a tab space in the name, and I'm unsure if it should be included or not~

Looks like uploads work with or without the trailing whitespace, so I've omitted it here.
Screen Shot 2022-07-12 at 11 04 54 AM
.


igor47:
we probably want to support the older templates that are circulating out there, too. we can make the list canonical, but now those older sheets will fail validation. i suggest also relaxing the validation criteria -- here, maybe it means just ignoring `'` and `,` characters, like we do with drop-list validation in https://github.com/usdigitalresponse/arpa-reporter/blob/3c43cf1c230eba7f03c1f3709e4b744e4521512e/src/server/services/validation-rules.js#L36-L38 ?

not sure what to do about the lead one, maybe we can keep both strings in the code somehow? or maybe we don't even care about the string, and we can just extract the numeric code at the beginning and that's okay?

trumanc:
This is further complicated by not knowing exactly what values the bulk upload endpoint will allow ultimately. Since we almost certainly won't be able to keep the input templates exactly correct (and users will likely be using old/out-of-date templates anyways), Joe has suggested in https://github.com/usdigitalresponse/arpa-reporter/issues/419#issuecomment-1182639779 that the users might just have to send the files to admins to correct EC codes when something is wrong.
With that in mind:

I've relaxed the validation by changing the error to be a warning if the code name doesn't exactly match. It also includes some additional messaging that the code might need to be corrected for full submission, even if it doesn't fail the current validation. This will allow people to continue to use old templates, but at least get some warning that it being out-of-date might cause them issues further down.

ajhyndman:
I think this is an improvement, and I'm happy to land it!

I actually think we could probably get away with only verifying that the numeric code matches a known code, and ignoring any other mismatch.

Then, in the report generation stage, we can map the EC code back to a single source of truth string that is defined in code.

ajhyndman (approved):

trumanc:
Yeah, I was thinking it would be nice to just do the mapping when generating output. My biggest worry is that if our source-of-truth is off then it would be pretty annoying if we actually altered someone's correct input to be something incorrect...

PR #425, opened 2022-07-13
Format subcategories in treasury export

ajhyndman:
Rewrite all "detailed expenditure category" strings before printing to treasury output.

This ensures that regardless of input template used, we always output the current value expected by treasury.

### DIFF
--- a/src/server/lib/format.js
+++ b/src/server/lib/format.js
@@ -6,6 +6,7 @@
  * reliably succeed if uploads have all passed validation.
  */
 const round = require('lodash/round')
+const { ecCodes } = require('./arpa-ec-codes')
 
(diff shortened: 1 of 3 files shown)


trumanc (approved):
I was able to confirm this works locally!

I think your branch doesn't include my recent change (https://github.com/usdigitalresponse/arpa-reporter/pull/422/files#diff-b7d727de95d9999a70cc8e40eb2e0b083d82c2ca24fe75d54e8a5ed64303d144R64) so it's a little unclear how they will interact. It might be safer to rebase this PR and test again.

That would also give you a chance to update the warning message from my PR. Instead of saying "you might need to change this later" we should indicate that we will automatically update it to the new value in the bulkupload csvs when they are exported later.

PR #430, opened 2022-07-13
Release 2022-07-13

trumanc:


igor47 (approved):
lgtm. what's up with the merge commit from main into staging? did you push that to staging? i would prefer to rebase on main, but it should be a no-op in any case, since staging should always be up to date with main.

trumanc:
There was some merge conflict because the line changed in https://github.com/usdigitalresponse/arpa-reporter/pull/418/files has apparently been changed a couple times. I resolved the conflict to be the updated code (with forceQuotes enabled) and that automerged that blank merge commit for some reason?

PR #433, opened 2022-07-14
Don't bother validating description of EC codes

ajhyndman:
@trumanc I decided to split this off as a distinct PR so we can discuss the right approach separately.

I think it's probably less confusing to agency users to not indicate a warning at all as long as the numeric code matches a known subcategory.  How does that sound?

### DIFF
--- a/src/server/services/validate-upload.js
+++ b/src/server/services/validate-upload.js
@@ -53,7 +53,6 @@ async function validateEcCode ({ upload, records }) {
 
   const codeParts = codeString.split('-')
   const code = codeParts[0]
-  const desc = codeParts.slice(1, codeParts.length).join('-')
 
   if (!ecCodes[code]) {
     return new ValidationError(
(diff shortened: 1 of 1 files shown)


trumanc:
@ajhyndman 
Yeah, good call splitting it.

I still think there is good reason to give this warning here, and it is the combination of two facts:
1. We're taking it upon ourselves to alter the input provided by the user, to the thing that we *believe* is correct.
2. Because the source-of-truth is so uncertain/unreliable, we aren't actually 100% sure this is the correct value.

In the really unfortunate case where we have changed someone's input from a correct value to an incorrect one, it could be really helpful to the user to be explicitly notified about the change we made. Otherwise they might be very confused by an error message later when they upload to the treasury.

I think that advantage outweighs the disadvantage of "the user has to see an extraneous warning message".

ajhyndman:
You make some good points, @trumanc!

I think I still like this approach though.  My understanding is that the users who will have to deal with validation in the treasury portal are the state admins, whereas the users who will see our validation warnings are the individual agency users.

Since these are different people, I suspect that means that talking about the export and upload to treasury process won't actually mean anything to the agency users.  Similarly, the state admins won't necessarily have seen those validation warnings before they generate the treasury export.

trumanc (approved):
ah, that's a good point. I'm still wrapping my head around who all the various users are and what each user in particular needs.

LGTM

ajhyndman:
We can definitely iterate as we get feedback as well!

Issue #434, opened 2022-07-14
Expenditures pulling incorrectly into Individuals and Payments LT50K Uploads


jmartin-oh:
We identified an issue where it seems like Column E of the Payments to Individuals and expenditureLT50k outputs do not have the correct information in them based on the uploaded spreadsheet, it seems to be repeating the obligation amount. 

Here is what was reported (upload [aaf0edeb](https://oh-arpa-reporter-web.onrender.com/uploads/aaf0edeb-f60e-4376-bd86-4159e2565860) ) and attached is what was generated

![image](https://user-images.githubusercontent.com/107556199/179072146-03ef060a-d562-49fd-8b0a-0f7ee6fba60c.png)
[expenditureLT50K.xlsx](https://github.com/usdigitalresponse/arpa-reporter/files/9114937/expenditureLT50K.xlsx)
[payments to individuals.xlsx](https://github.com/usdigitalresponse/arpa-reporter/files/9114938/payments.to.individuals.xlsx)

PR #436, opened 2022-07-14
Fix Issue #434: Correct typo where Obligation was incorrectly used twice

trumanc:
Fix for https://github.com/usdigitalresponse/arpa-reporter/issues/434

The obligation field was incorrectly specified twice in a row, instead of giving the obligation and then the expenditures


ajhyndman (commented):
Oof, thank you for spotting that!

ajhyndman (approved):

PR #440, opened 2022-07-16
Knex returns the raw date string for date fields

### LINKED ISSUE #381: Date mismatch
Low priority at this time as there are not extensive date validations. :)

The Start and End dates do not match when comparing the Reporting Period "home" to each specific period (seen when editing). Even if if I edit to change it, it reverts back. 

image

image

ajhyndman:
Fixes #381 

It looks like @mattbroussard did correctly diagnose [a similar issue](https://github.com/usdigitalresponse/usdr-gost/issues/77) in the USDR GOST project.

What's happening is that when reading dates out of a date column in Postgres, knex passes the raw date string to a JS date constructor (e.g. `new Date('2022-09-30')`).  This resolves the date to an exact time (midnight on that date in the _server's_ timezone).  Then, when we serialize that date as JSON to send over the network, the time and timezone information are preserved.

Since the server's timezone and the user's timezone don't always agree, the date can end up +/- one day from when it was intended to be.

Rather than try to accommodate for server's timezone somewhere, we can prevent the addition of time information to Postgres dates by overriding the knex date parser.

BEFORE

image

AFTER

image


src/server/db/connection.js
```diff
@@ -1,6 +1,12 @@
+const { types } = require('pg')
 const { POSTGRES_URL } = require('../environment')
 
 console.log('\nConnecting to database:', POSTGRES_URL)
+
+// override parser for date fields — just return the raw string content
+const DATE_OID = 1082
```
igor47, line 7: where did the magic number come from?

igor47 (approved):
lgtm, i wonder if the constant is available somewhere inside the `pg` library, or maybe add a comment explaining where you found the constant?

src/server/db/connection.js
ajhyndman: I followed Matt's research above to find this reference.

It looks like the constant is available via a separate package, `pg-types`.  I wasn't sure it was worth installing for a single constant, but we certainly can.  🤷

PR #447, opened 2022-07-18
Revert "Read project demographics options from dropdowns sheet"

trumanc:
Reverts usdigitalresponse/arpa-reporter#439

This merge has caused errors on staging. I was able to repro the bug locally, and then fix it by reverting this change. Here's the error message from the logs on staging:

Screen Shot 2022-07-18 at 3 42 13 PM


ajhyndman:
Ooph, thanks for responding to the issue quickly.  Let me see if I can find my mistake.

PR #464, opened 2022-07-22
Update reference to Small_Businesses_Served__c

### LINKED ISSUE #463: Zeroes are dropped from project bulk upload CSVs in treasury export
We should preserve zeroes when generating our output.

ajhyndman:
It looks like I overlooked this field name in the June 21st template updates!

Fixes #463

### DIFF
--- a/src/server/services/generate-arpa-report.js
+++ b/src/server/services/generate-arpa-report.js
@@ -153,7 +153,7 @@ async function generateProject18 (records) {
             record.content.Tertiary_Proj_Demographics_Explanation__c,
             record.content.Structure_Objectives_of_Asst_Programs__c,
             record.content.Recipient_Approach_Description__c,
-            record.content.Number_Non_Profits_Served__c
+            record.content.Small_Businesses_Served__c ?? record.content.Number_Non_Profits_Served__c
           ]
         }
         default:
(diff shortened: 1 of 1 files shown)


src/server/services/generate-arpa-report.js
```diff
@@ -199,7 +199,7 @@ async function generateProject19 (records) {
             record.content.Tertiary_Proj_Demographics_Explanation__c,
             record.content.Structure_Objectives_of_Asst_Programs__c,
             record.content.Recipient_Approach_Description__c,
-            record.content.Number_Non_Profits_Served__c
+            record.content.Small_Businesses_Served__c ?? record.content.Number_Non_Profits_Served__c
```
trumanc, line 202: EC 1.9 I think actually DOES want the number of Non-profits, and this line should be reverted 
Screen Shot 2022-07-22 at 8 55 55 AM
d

trumanc (approved):
I would double-check that change to 1.9, then this LGTM

src/server/services/generate-arpa-report.js
ajhyndman: Thanks, you're definitely right!

In fact, it looks like I specifically changed this field in 1.8 to the wrong value when making the last round of updates.  I guess that measn we shouldn't need to keep the fallback lookup in template 1.8 either.

I wonder how I made that mistake.

trumanc (approved):

PR #468, opened 2022-07-22
Extract dropdown values from correct tab

trumanc:
The `generateRules` script was still pulling dropdown value lists from the hidden row within each tab that had inconsistent/incorrect values in many cases. 

We had identified and corrected some of those mistakes manually in `validation-rules.json`, but hadn't addresses the issue systematically so we were regularly getting more errors reported. 

This PR changes the system to ignore those often-incorrect lists, and instead uses the lists recorded in the Dropdowns tab of the worksheet to populate the extracted rules. The change resulted in around 370 lines of corrected validations.

It requires manually recording which field-ids map to which dropdown names, so I've added that mapping to the script, and included an error check so that when the Dropdowns tab is updated in the future it will remind us to update the mappings at the same time.

NOTE: This structure assumes that every instance of a certain field-id will adhere to the same list of values. If that assumption ever changes, then we'll have to update this system to include EC codes in the mappings.

### DIFF
--- a/src/server/services/validation-rules.js
+++ b/src/server/services/validation-rules.js
@@ -1,6 +1,5 @@
 
 const srcRules = require('../lib/templateRules.json')
-const srcDropdowns = require('../lib/templateDropdowns.json')
 
 const recordValueFormatters = {
   makeString: (val) => String(val),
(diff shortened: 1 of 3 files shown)


src/server/services/validation-rules.js
```diff
@@ -1,6 +1,5 @@
 
 const srcRules = require('../lib/templateRules.json')
-const srcDropdowns = require('../lib/templateDropdowns.json')
```
igor47, line 3: we might be safe to remove this file from the repo now

igor47 (approved):
beautiful

src/server/services/validation-rules.js
trumanc: ooh I hadn't considered that. Yeah, it doesn't get used anywhere else now, so it's definitely safe.

I kind of like it though, because it gives us a git history of changes to dropdown values in a way that is much easier to see than if we had to open up all the old worksheets and inspect them manually.

ajhyndman:
Oh, thanks so much for taking this on!

PR #469, opened 2022-07-22
Correct template filename typo

trumanc:
Noticed a filename was incorrect, and verified that it caused errors when trying to export data of those project types. After correction the export worked correctly.


ajhyndman (approved):
Whoops, thank you!

PR #471, opened 2022-07-23
Collect projects from all reporting periods in audit report

### LINKED ISSUE #432: Audit Report: Project Summaries tab should list ALL projects
The Project Summaries tab should list every project code that has ever been validated exactly once.

Bonus: Add a new column that reports which period the data comes from.

ajhyndman:
Collect the latest version of each project in the audit report's project summaries tab.

Fixes #432

### DIFF
--- a/src/server/db/reporting-periods.js
+++ b/src/server/db/reporting-periods.js
@@ -30,6 +30,7 @@ module.exports = {
   getReportingPeriod,
   closeReportingPeriod,
   getReportingPeriodID,
+  getPreviousReportingPeriods,
   getAllReportingPeriods,
   createReportingPeriod,
   updateReportingPeriod
@@ -78,6 +79,23 @@ async function getReportingPeriodID (tenantId, periodID) {
   requiredArgument(tenantId, 'must specify tenantId in getReportingPeriodID')
 
   return Number(periodID) || getCurrentReportingPeriodID(tenantId)
+}
+
+/**
+ * Get all reporting periods that either match supplied poriod ID or are older
+ * than supplied period ID.
+ *
+ * @returns The matching reporting periods, sorted from oldest to newest by date
+ */
+async function getPreviousReportingPeriods (tenantId, period_id, trns = knex) {
+  const currentReportingPeriod = await getReportingPeriod(tenantId, period_id)
+  const allReportingPeriods = await getAllReportingPeriods(tenantId)
+  const reportingPeriods = allReportingPeriods.filter(
+    period =>
+      new Date(period.end_date) <= new Date(currentReportingPeriod.end_date)
+  )
+  reportingPeriods.sort((a, b) => new Date(a.end_date) - new Date(b.end_date))
+  return reportingPeriods
 }
 
 async function closeReportingPeriod (user, period, trns = knex) {
--- a/src/server/db/reporting-periods.js
+++ b/src/server/db/reporting-periods.js
@@ -82,7 +82,7 @@ async function getReportingPeriodID (tenantId, periodID) {
 }
 
 /**
- * Get all reporting periods that either match supplied poriod ID or are older
+ * Get all reporting periods that either match supplied period ID or are older
  * than supplied period ID.
  *
  * @returns The matching reporting periods, sorted from oldest to newest by date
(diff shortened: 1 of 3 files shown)


src/server/db/reporting-periods.js
```diff
@@ -80,6 +81,23 @@ async function getReportingPeriodID (tenantId, periodID) {
   return Number(periodID) || getCurrentReportingPeriodID(tenantId)
 }
 
+/**
+ * Get all reporting periods that either match supplied poriod ID or are older
```
trumanc, line 85: nit: typo here
```suggestion
 * Get all reporting periods that either match supplied period ID or are older
```

src/server/services/records.js
```diff
@@ -110,6 +114,7 @@ async function recordsForUpload (upload) {
 }
 
 async function recordsForReportingPeriod (tenantId, periodId) {
+  log('recordsForReportingPeriod()')
```
trumanc, line 117: Did you mean to include this log?

src/server/services/records.js
```diff
@@ -120,9 +125,43 @@ async function recordsForReportingPeriod (tenantId, periodId) {
   return groupedRecords.flat()
 }
 
+/**
+ * Get the most recent, validated record for each unique project, as of the
+ * specified reporting period.
+*/
+async function mostRecentProjectRecords (tenantId, periodId) {
+  log('mostRecentProjectRecords()')
```
trumanc, line 133: and this one?

src/server/services/records.js
```diff
@@ -120,9 +125,43 @@ async function recordsForReportingPeriod (tenantId, periodId) {
...
+  )
+
+  const latestProjectRecords = allRecords
+    .flat()
+    // exclude non-project records
+    .filter(record => Object.values(EC_SHEET_TYPES).includes(record.type))
+    // collect the latest record for each project ID
+    .reduce(
+      (accumulator, record) => ({
+        ...accumulator,
```
trumanc, line 152: I'm not super familiar with the js spread operator, but is it right to assume that each time this reduce callback is called it has to create a whole new object copied from the previous object, add/overwrite an entry, and then return it? Would that make this an O(N^2) iteration, instead of just O(N) if you did this instead?
```
.reduce(
  (accumulator, record) => {
    accumulator[record.content.Project_Identification_Number__c] = record
    return accumulator
})
```

trumanc (approved):
LGTM!

I'm a little nervous about how this code will scale. Generating the audit report now has to read every single record for the tenant *twice*. I don't think that's an issue yet, but as there are more tenants (so this code is called more often) and more reporting periods/records (so each call is more expensive) this may eventually need to be optimized. Not a priority for now, but something to consider in the future.

src/server/services/records.js
ajhyndman: I did!  We have used this pattern in a few places.  The `log` function is a no-op unless the `VERBOSE` environment variable is set, and logging the name of the function calls can be helpful debugging.  🤷

src/server/services/records.js
ajhyndman: Yes, this is an accurate take!

However, in my experience, the cost of object manipulation in time and memory is negligible compared to performance costs like the I/O issue you identified above.

I like to default to this spread pattern just because it avoids any possibility of accidental non-local mutation.

To be fair, I don't think either accidental mutation or performance overhead are likely to be meaningful here.  😅

ajhyndman:
> I'm a little nervous about how this code will scale. Generating the audit report now has to read every single record for the tenant _twice_. I don't think that's an issue yet, but as there are more tenants (so this code is called more often) and more reporting periods/records (so each call is more expensive) this may eventually need to be optimized. Not a priority for now, but something to consider in the future.

You make a very astute observation.  I had a similar thought while writing the code, but I guess I was too lazy to try to solve it immediately, haha.

My first idea for how to optimize this was be to memoize `getRecordsForUpload`.  That way the calling code doesn't have to worry about performance, but subsequent reads for the same upload won't hit the filesystem.  The question mark here though is how long to hold onto the cache.  I don't think we want the node process to retain all records in memory indefinitely.  That would effectively become a memory leak and could eventually bog down the server.

Anyway, that's about as far as I got before I decided to worry about it later.

PR #472, opened 2022-07-23
Add persistentFormatters and dropdown corrections

trumanc:
Adds two new systems:
1) a new system for doing value formatting. These validations are more powerful (and dangerous) than the previous formatting, because they will also affect the values that are printed to exports.

2) A structured system for recording problems we've detected with dropdown values. It lets us easily refer to a specific value in the worksheet, and provide information about what that value should be corrected to, and what other values we should allow. This system uses the new formatting system to coerce any allowable values into the correct one.

(also fixes a bug I found where 1.9 projects were not exporting the cancellation reason)

Here are some output csvs, shown pre-fix and post-fix to show how the output dropdown values were fixed. Both cases used the same input uploads, which included the old dropdown values as their value.
The post-fix export csvs correctly show the new dropdown values, with their newly enforced commas and space.

Projects:
Screen Shot 2022-07-23 at 1 40 52 PM

Awards:
Screen Shot 2022-07-23 at 1 42 37 PM

### DIFF
--- a/src/server/services/generate-arpa-report.js
+++ b/src/server/services/generate-arpa-report.js
@@ -179,6 +179,7 @@ async function generateProject19 (records) {
             record.content.Name,
             record.content.Project_Identification_Number__c,
             record.content.Completion_Status__c,
+            record.content.Cancellation_Reason__c,
             currency(record.content.Adopted_Budget__c),
             currency(record.content.Total_Obligations__c),
             currency(record.content.Total_Expenditures__c),
--- a/src/server/services/records.js
+++ b/src/server/services/records.js
@@ -80,8 +80,9 @@ async function recordsForUpload (upload) {
       continue
     }
 
+    const typeRules = rules[type]
     // header is based on the columns we have in the rules
-    const header = Object.values(rules[type])
+    const header = Object.values(typeRules)
       .sort((a, b) => a.index - b.index)
       .map(rule => rule.key)
 
@@ -102,7 +103,19 @@ async function recordsForUpload (upload) {
 
     // each row in the input sheet becomes a unique record
     for (const row of rows) {
-      records.push({ type, subcategory, upload, content: row })
+      const formattedRow = {}
+      Object.keys(row).forEach(fieldId => {
+        let value = row[fieldId]
+        for (const formatter of typeRules[fieldId].persistentFormatters) {
+          try {
+            value = formatter(value)
+          } catch (e) {
+            console.log(`Persistent formatter failed to format value ${value} with error:`, e)
+          }
+        }
+        formattedRow[fieldId] = value
+      })
+      records.push({ type, subcategory, upload, content: formattedRow })
     }
   }
 
--- a/src/server/services/records.js
+++ b/src/server/services/records.js
@@ -80,9 +80,9 @@ async function recordsForUpload (upload) {
       continue
     }
 
-    const typeRules = rules[type]
+    const rulesForCurrentType = rules[type]
     // header is based on the columns we have in the rules
-    const header = Object.values(typeRules)
+    const header = Object.values(rulesForCurrentType)
       .sort((a, b) => a.index - b.index)
       .map(rule => rule.key)
 
@@ -106,7 +106,7 @@ async function recordsForUpload (upload) {
       const formattedRow = {}
       Object.keys(row).forEach(fieldId => {
         let value = row[fieldId]
-        for (const formatter of typeRules[fieldId].persistentFormatters) {
+        for (const formatter of rulesForCurrentType[fieldId].persistentFormatters) {
           try {
             value = formatter(value)
           } catch (e) {
--- a/src/server/services/validation-rules.js
+++ b/src/server/services/validation-rules.js
@@ -3,7 +3,7 @@ const srcRules = require('../lib/templateRules.json')
 
 const recordValueFormatters = {
   makeString: (val) => String(val),
-  trimWhitespace: (val) => val.trim(),
+  trimWhitespace: (val) => val ? val.trim() : val,
   removeCommas: (val) => val.replace(/,/g, ''),
   removeSepDashes: (val) => val.replace(/^-/, '').replace(/;\s*-/g, ';'),
   toLowerCase: (val) => val.toLowerCase()
@@ -12,7 +12,7 @@ const recordValueFormatters = {
 /*
 Structured data recording all the immediate corrections we want to apply to dropdowns.
 There are 2 types of corrections we can apply:
-1) The value in the current worksheet is incorrect.
+1) The value in the currently committed input template is incorrect.
 2) A value in the dropdown list changed in the past, and we want to continue to allow legacy vals as valid inputs.
 In the first case, we will alter the validation rule to check against the new correct value, and
 then treat the value currently seen in the worksheet as an allowable legacy value.
@@ -61,7 +61,7 @@ function generateRules () {
 
       if (rule.dataType === 'String') {
         rule.validationFormatters.push(recordValueFormatters.makeString)
-        rule.validationFormatters.push(recordValueFormatters.trimWhitespace)
+        rule.persistentFormatters.push(recordValueFormatters.trimWhitespace)
       }
 
       if (rule.dataType === 'Multi-Select') {
--- a/src/server/services/validation-rules.js
+++ b/src/server/services/validation-rules.js
@@ -6,8 +6,35 @@ const recordValueFormatters = {
   trimWhitespace: (val) => val.trim(),
   removeCommas: (val) => val.replace(/,/g, ''),
   removeSepDashes: (val) => val.replace(/^-/, '').replace(/;\s*-/g, ';'),
-  toLowerCase: (val) => val.toLowerCase(),
-  childCareSpacing: (val) => (val === 'family or childcare' ? 'family or child care' : val)
+  toLowerCase: (val) => val.toLowerCase()
+}
+
+/*
+Structured data recording all the immediate corrections we want to apply to dropdowns.
+There are 2 types of corrections we can apply:
+1) The value in the current worksheet is incorrect.
+2) A value in the dropdown list changed in the past, and we want to continue to allow legacy vals as valid inputs.
+In the first case, we will alter the validation rule to check against the new correct value, and
+then treat the value currently seen in the worksheet as an allowable legacy value.
+In both cases, we will foribly coerce any instances of legacy values into the correct value.
+This coercion happens whenever we read the value from the upload file, so it will apply to
+validations we perform, as well as values we export in reports.
+   */
+const dropdownCorrections = {
+  'Affordable housing supportive housing or recovery housing': {
+    correctedValue: 'Affordable housing, supportive housing, or recovery housing'
+  },
+  'COVID-19 testing sites and laboratories and acquisition of related equipment': {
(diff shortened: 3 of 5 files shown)


ajhyndman (commented):
This is looking pretty good!  I have a couple high level thoughts about approach, but getting this working is relatively urgent, and refactoring later is always an option.

---

First, I wonder if we can get away with just applying these transforms to all fields.  It feels to me like we're picking up a fair amount of complexity trying to preserve and match information about field types to the application of our formatters.

If we can assume that if a field anywhere has content that exactly matches one of our known misspellings, it seems pretty safe to assume that is unintentional?

I think this would also let us move the transformation from record extraction to export generation, which puts it more in line with the other formatters we are applying.

e.g.

```js
const escapedContent = [...template, ...csvData].map(row =>
  row
    .map(normalizeDropdowns)
    // 2022-07-08 Treasury expects LF line endings _within_ cells and CRLF line
    // endings _between_ rows.  Failure to adhere to either of these requirements
    // can result in an opaque error that looks like the following:
    // "List index out of bounds: 1"
    .map(value => typeof value === 'string' ? value.replace(/\r\n/g, '\n') : value)
)
```

---

Second, if we don't think we can get away with the above, I also have a bias against coupling ourselves too closely to the Excel input template.  I'm worried that the update cycle for input templates can be slow, because it requires multiple people's involvement and a whole email-based distribution.

Conversely, any validations or formatting we can specify directly in code only requires a single engineer to update.

I think if it were just up to me, I'd try to implement the normalization as just another transform in `src/server/lib/formatters` and apply it manually to any known dropdown (or multiselect) field.

I also like being able to scan this file and understand exactly what fields are mapped where and which transforms are being applied, even if it is a little tedious to write.

---

I think @igor47 might disagree with me on these positions, so feel free to push back!

trumanc:
For your first point:
I considered that, but decided to go with this more complex setup to protect it against a (probably unlikely but possible) error case that treasury might have. That case is: if Treasury ever uses one version of a string constant somewhere, but require a different version elsewhere, then this blanket "apply the formatting everywhere" problem will not be able to fix it, and our hands are tied. This more granular approach could adapt to handle that more easily.

Also, by introducing this new formatting system *earlier* in the process (at record read) instead of *later* (at export writing) it means that when we move formatters into this system, we can delete them from the validation system (notice that I was able to delete the "childcareSpacing" special case formatter, because it is handled by the dropdown correction now)


For point 2:
The point of this system is that it lets us make these corrections WITHOUT updating the template. It lets us handle these errors in the input template without actually having to update the template (and without making the admins correct dropdown values in the template, like Ozzie has been doing).
But because the system is made to fix issues currently in the template, it is coupled in that when we DO eventually fix the template, it will throw an error and make sure you've cleaned up old corrections that are no longer relevant.


> I think if it were just up to me, I'd try to implement the normalization as just another transform in src/server/lib/formatters and apply it manually to any known dropdown (or multiselect) field.

My complaint with that is that we would have to add the transform to EVERY dropdown, and there are tons of them. I would almost guarantee we miss one, or forget to add it when we add a new dropdown field later.

src/server/services/generate-arpa-report.js
```diff
@@ -179,6 +179,7 @@ async function generateProject19 (records) {
             record.content.Name,
             record.content.Project_Identification_Number__c,
             record.content.Completion_Status__c,
+            record.content.Cancellation_Reason__c,
```
trumanc, line 182: Found this bug while I was testing the output: Notice in the output screenshots that the pre-fix 1.9 projects report has all the data shifted over one column, but it's corrected in the post-fix screenshot

src/server/services/validation-rules.js
```diff
@@ -6,8 +6,35 @@ const recordValueFormatters = {
...
+const dropdownCorrections = {
+  'Affordable housing supportive housing or recovery housing': {
+    correctedValue: 'Affordable housing, supportive housing, or recovery housing'
+  },
+  'COVID-19 testing sites and laboratories and acquisition of related equipment': {
+    correctedValue: 'COVID-19 testing sites and laboratories, and acquisition of related equipment'
+  },
+  'Family or child care': {
+    allowableLegacyValues: ['Family or childcare']
+  }
```
ajhyndman, line 32: Do you still need this reverse lookup if we perform formatting before validation?

ajhyndman:
> if Treasury ever uses one version of a string constant somewhere, but require a different version elsewhere, then this blanket "apply the formatting everywhere" problem will not be able to fix it, and our hands are tied

I guess my preferences would still be to stay simple until complex becomes necessary, but I don't feel that strongly.

> Also, by introducing this new formatting system earlier in the process (at record read) instead of later (at export writing) it means that when we move formatters into this system, we can delete them from the validation system (notice that I was able to delete the "childcareSpacing" special case formatter, because it is handled by the dropdown correction now)

Oh, this is a neat advantage.  I hadn't thought of that.

> The point of this system is that it lets us make these corrections WITHOUT updating the template.

The coupling I'm talking about is the way that we are deriving field type in order to decide which formatter to apply.  Unless I'm misunderstanding, that still has to come from the input template, correct?

> My complaint with that is that we would have to add the transform to EVERY dropdown, and there are tons of them. I would almost guarantee we miss one, or forget to add it when we add a new dropdown field later.

This is certainly fair, but I also think it'll be easy to spot and quick to fix when it happens.

src/server/services/validation-rules.js
trumanc: Structuring the lookup this way makes it easier to do a couple things:
1) We can easily verify that the rule is configured to mention a real value in a dropdown somewhere (which I added in `validateDropdownCorrections` in the generate-rules script)
2) It lets us easily add new legacy values if something changes again in the future. E.g. It seems like at some point in the past they changed their capitalization strategy, and then they changed their commas strategy, and then they changed the commas back to the old system. By indexing by the current value and keeping a list of all legacy values, it's easy to add those corrections to the current config.

trumanc:
> The coupling I'm talking about is the way that we are deriving field type in order to decide which formatter to apply. Unless I'm misunderstanding, that still has to come from the input template, correct?

I think this change is actually less coupled to the input template than you might think.
This new formatter is triggered when the listVals for a rule includes a value listed in the dropdownCorrections object, but those listVals are less dependent on the template than ever. My recent PR (https://github.com/usdigitalresponse/arpa-reporter/pull/468/files#diff-d124346f5eb07033826d3c4d46726717eb3a5ed1ab3b1b481558fe357454ba36R19) updated the listVal processing to rely on this in-code mapping of field-id to dropdownName instead of relying on whether the template correctly marked a column as "Pick List".

That processing does still rely on the actual dropdowns listed in the template's "Dropdowns" tab, but the whole point of this change is that it allows us to say "if something in the Dropdowns tab was incorrect, we can correct it in code now".


So unless I'm misunderstanding something, I think this PR actually moves us towards de-coupling from the template, and makes it easier to fix things faster in code instead.

src/server/services/validation-rules.js
```diff
@@ -25,24 +52,37 @@ function generateRules () {
...
+      // persistentFormatters are always applied as soon as a value is read from a
+      // an upload, so they will affect both validation AND the exported value.
+      rule.validationFormatters = []
+      rule.persistentFormatters = []
 
       if (rule.dataType === 'String') {
-        rule.valueFormatters.push(recordValueFormatters.makeString)
-        rule.valueFormatters.push(recordValueFormatters.trimWhitespace)
+        rule.validationFormatters.push(recordValueFormatters.makeString)
+        rule.validationFormatters.push(recordValueFormatters.trimWhitespace)
```
trumanc, line 64: This would fix the "whitespace appearing in email addresses" issue that just came up

```suggestion
        rule.persistentFormatters.push(recordValueFormatters.trimWhitespace)
```

src/server/services/records.js
```diff
@@ -80,8 +80,9 @@ async function recordsForUpload (upload) {
       continue
     }
 
+    const typeRules = rules[type]
```
ajhyndman, line 83: Nit: It took me a second to figure out what this variable was.

Are you comfortable with a more verbose name, like `rulesForCurrentType`?

src/server/services/records.js
```diff
@@ -102,7 +103,19 @@ async function recordsForUpload (upload) {
...
+      Object.keys(row).forEach(fieldId => {
+        let value = row[fieldId]
+        for (const formatter of typeRules[fieldId].persistentFormatters) {
+          try {
+            value = formatter(value)
+          } catch (e) {
+            console.log(`Persistent formatter failed to format value ${value} with error:`, e)
+          }
+        }
+        formattedRow[fieldId] = value
```
ajhyndman, line 116: Totally optional, but since we have lodash in the project, you could try leveraging the `_.flow` pattern here!

```suggestion
try {
  const value = _.flow(typeRules[fieldId].persistentFormatters)(row[fieldId])
  formattedRow[fieldId] = value
} catch (e) {
  console.log(`Persistent formatter failed to format value ${value} with error:`, e)
}
```

I guess one big difference is that your approach supports skipping an individual broken formatter.  I'm not sure if that's likely to be more or less confusing in practice.

src/server/services/validation-rules.js
```diff
@@ -25,24 +52,37 @@ function generateRules () {
   // for any values we format, we should format them the same way when we export
   for (const ruleType of Object.keys(rules)) {
     for (const rule of Object.values(rules[ruleType])) {
-      rule.valueFormatters = []
+      // validationFormatters are only applied when validating the records, so they
+      // aren't used during exports.
+      // persistentFormatters are always applied as soon as a value is read from a
+      // an upload, so they will affect both validation AND the exported value.
```
ajhyndman, line 58: This comment is very helpful, thanks!

src/server/services/validation-rules.js
ajhyndman: Oh!   You're working around the fact that the latest _input template_ that Joe published still has values that aren't corrected?

Sorry, this took me a long time to grok.

Oof, that's a bit of a pain.

I'd be a lot happier if we only had one format for specifying corrections, but I guess we'd have to sacrifice the validation you wrote to achieve that?

> 1) The value in the current worksheet is incorrect.

If I've got it now, this is referring to the currently committed input template, not the worksheet that you are currently processing.  Is that right?

I'm happy to worry about refactoring later, but (if I have it right) would you mind updating this comment for clarity?

> 1) The value in the currently committed input template is incorrect.

ajhyndman (approved):
@trumanc Could you just double-check my understanding in [this comment](https://github.com/usdigitalresponse/arpa-reporter/pull/472#discussion_r929476655)?

Assuming I'm right, my only blocking request is that we update that comment.

I left a couple other super minor suggestions, but the rest looks good!

src/server/services/records.js
trumanc: yeah, good call. I'll update this now

src/server/services/records.js
trumanc: I do like how the `flow` code reads, but I think I prefer the functionality of not short-circuiting if one is broken. Mostly because this way we know that any broken formatter will write an error message out (which will be even more useful once we get sentry hooked up)

src/server/services/validation-rules.js
trumanc: yes, that's correct! I'll update the comment now to reference the currently committed input template

ajhyndman (approved):

PR #497, opened 2022-08-21
Memoize filesystem reads for uploads (within request)

ajhyndman:
Add a memoization layer to upload reads.  This ensures that the same upload is not read from disk more than once within a single request.

The audit report generation process could end up reading the same upload from disk two or more times.  Rather than try to complicate the logic for report generation, this change caches audit report reads.

To avoid leaking memory, this cache is only preserved on the express request object.  This allows the node process to garbage collect cached uploads after the request has resolved.  This will be very important in a multi-tenant deployment.

### DIFF
--- a/src/server/services/records.js
+++ b/src/server/services/records.js
@@ -150,10 +150,10 @@ async function recordsForUpload (upload) {
     req.recordsForUpload = {}
   }
   if (req.recordsForUpload[upload.id]) {
-    log(`recordsForUpload(${upload.id}): waiting for read`)
+    log(`recordsForUpload(${upload.id}): reading from cache`)
     return req.recordsForUpload[upload.id]
   }
-  log(`recordsForUpload(${upload.id}): fetching`)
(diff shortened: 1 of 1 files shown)


src/server/services/records.js
```diff
@@ -126,8 +134,37 @@ async function recordsForUpload (upload) {
...
+ */
+async function recordsForUpload (upload) {
+  log(`recordsForUpload(${upload.id})`)
+
+  const req = useRequest()
+  if (!req.recordsForUpload) {
+    req.recordsForUpload = {}
+  }
+  if (req.recordsForUpload[upload.id]) {
+    log(`recordsForUpload(${upload.id}): waiting for read`)
```
mattbroussard, line 153: nit: this log isn't quite accurate; we don't know whether the promise has resolved or not when we return it here

mattbroussard (approved):

src/server/services/records.js
ajhyndman: Oh yeah, thanks!

PR #504, opened 2022-08-26
Fix format.spec.js test broken by 3ec63f4ceb6

mattbroussard:
Test was broken on `staging` branch, caused by 3ec63f4ceb6


ajhyndman (approved):
Thanks for the fix

Issue #487, opened 2022-08-11
Generate Audit Report is too slow


ajhyndman:
Can we optimize the processing of this report?

mattbroussard:
#497 is an optimization that helps this.

If we get to a point where audit report generation is complicated enough that optimizations like that don't get us below the request timeout (do we know what that limit is?) we could consider decoupling report generation from request lifetime -- e.g. the client sends one request to start generation, which kicks off the process and returns immediately; then the client polls in separate requests to find out when the job is done.

mindyatwork:
Closing for now, will ask partners to resubmit issue if it's still too slow.

PR #509, opened 2022-09-01
Collate report data more efficiently

trumanc:
Quick update that should help with audit report slowness as reported in https://github.com/usdigitalresponse/arpa-reporter/issues/487 by changing some O(N) object-creation operations to O(1) object-update operations.

The splat operator is causing us the code to create a new object everytime, instead of just making small modifications to the object in place.

This is particularly troublesome in records.js, where the object has an entry for every single project for this tenant, and we are looping over every historical submission/version of each record.


ajhyndman (approved):

ajhyndman:
This is definitely worth a shot, thanks!

PR #510, opened 2022-09-01
GOST integration: misc preparations for copy

mattbroussard:
- Remove some bad syntax that Babel (in my [import rewriter](https://github.com/usdigitalresponse/usdr-gost/pull/297) script) choked on
- add an environment variable check to gate VueRouter `base` on whether running in standalone ARPA reporter repo or GOST
- fix dependencies on being mounted at root path
  - remove raw reference to document.location in Navigation.vue; use VueRouter's `$route` object instead which accounts for varying `base`
  - replace some `a` tags with `router-link` tags
- organize `configure.js` into separate functions that can be more easily called from GOST's after repo merge

### DIFF
--- a/src/server/configure.js
+++ b/src/server/configure.js
@@ -10,13 +10,7 @@ const environment = require('./environment')
 const publicPath = resolve(__dirname, '../../dist')
 const staticConf = { maxAge: '1y', etag: false }
 
-module.exports = (app, options = {}) => {
-  if (!options.disableRequestLogging) {
-    app.use(morgan('common'))
-  }
-  app.use(bodyParser.json())
-  app.use(cookieParser(environment.COOKIE_SECRET))
-
+function configureApiRoutes (app) {
   app.use('/api/agencies', require('./routes/agencies'))
   app.use(
     '/api/application_settings',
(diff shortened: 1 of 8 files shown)


src/server/configure.js
```diff
@@ -10,13 +10,7 @@ const environment = require('./environment')
...
 
-module.exports = (app, options = {}) => {
-  if (!options.disableRequestLogging) {
-    app.use(morgan('common'))
-  }
-  app.use(bodyParser.json())
-  app.use(cookieParser(environment.COOKIE_SECRET))
-
+function configureApiRoutes (app) {
   app.use('/api/agencies', require('./routes/agencies'))
```
trumanc, line 14: I see that the baseurl has been updated in the Vue router, but do we need any similar update for the api endpoints in the express app? Are we sure there are no collisions between this and the other GOST api urls?

trumanc (approved):

src/server/configure.js
mattbroussard: Most GOST [routes](https://github.com/usdigitalresponse/usdr-gost/blob/_staging/packages/server/src/configure.js) are namespaced under `/api/organizations/: organizationId` so this won't be an issue. The only collision is `/api/sessions` and my other PR has an import rewrite so the reference to it here will import GOST's (and the app.use call will no-op anyway since requests to that path will be handled by an earlier middleware).

Sometime after merge we may consider namespacing ARPA routes for clarity but it's not a technical necessity.

ajhyndman (approved):

tests/server/fixtures/add-dummy-data.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
 if (require.main === module) {
   require('dotenv').config()
   const knex = require('../../../src/server/db/connection')
-  return setupAgencies(knex).then(() => {
+  setupAgencies(knex).then(() => {
```
ajhyndman, line 20: Wow, I wonder how this happened.

PR #513, opened 2022-09-06
Followup fix from #505: handle role properly when creating new users

mattbroussard:
This fixes one callsite missed in #505's rework of how user objects are dealt with: when creating a new user in legacy ARPA Reporter environment, need to convert `user.role` object back into string (in GOST, this code is replaced).


ajhyndman (approved):
👍

PR #515, opened 2022-09-07
GOST integration: ARPA-side changes to allow running ARPA tests in GOST repo

mattbroussard:
Sister PR to https://github.com/usdigitalresponse/usdr-gost/pull/310, this makes ARPA-side changes to allow ARPA tests to run in the GOST repo.

Namely:
- `SRC_DIR` is refactored into `SERVER_CODE_DIR` to remove dependency on a specific directory structure
  - its test is updated accordingly, as well as removing dynamic require which my import rewriter script cannot handle. I don't really see a strong reason it was like this before? Perhaps @ajhyndman or @igor47 have an idea?
- Add a seed for GOST's `tenants` table (no-ops in legacy ARPA repo)
- Modify `add-dummy-data` script to conditionally populate required `agencies.main_agency_id` column when running in GOST repo
- Modify test seeds for `roles` and `users` to support the GOST versions of those tables
- Adjust the regexes used in `requireSrc` to handle GOST's directory layout (also though about just replacing all of these with regular relative `require`s that could be handled by the import rewriter, but meh, there are a number of these)
- Test runner changes
  - Change where the `.env` file is found, conditionally
  - Handle environment variables whose name contains an `=` symbol correctly (e.g. `NODE_OPTIONS`)
  - Don't depend on specific directory structure when referencing tests and test seeds

There are still a few other things that need to be addressed to get tests to *pass*, notably adjusting the seed data like I did in my dev branch's https://github.com/usdigitalresponse/usdr-gost/pull/308/commits/b66ff2ff736306524b03d10b25ec08c7bd0f205f -- but I'm going to take a separate look at that in the context of other seed-related changes.

### DIFF
--- a/tests/server/environment.spec.js
+++ b/tests/server/environment.spec.js
@@ -1,14 +1,40 @@
 const assert = require('assert')
+const path = require('path')
+const fs = require('fs/promises')
 
-const underTest = '../../src/server/environment'
+const env = require('../../src/server/environment')
 
 describe('environment settings', function () {
-  beforeEach('clear env module', function () {
-    delete require.cache[require.resolve(underTest)]
+  it('SERVER_CODE_DIR point to the right place', async function () {
+    const testPaths = [
+      'environment.js',
+      'lib/ensure-async-context.js'
+    ]
+
+    for (const relative of testPaths) {
+      const absolute = path.join(env.SERVER_CODE_DIR, relative)
+      const pathExists = await exists(absolute)
+      assert.ok(pathExists, `expected ${absolute} to exist, is SERVER_CODE_DIR wrong?`)
+    }
   })
 
-  it('points the CODE_DIR at src', function () {
-    const env = require(underTest)
-    assert.equal(env.SRC_DIR.slice(env.SRC_DIR.length - 3), 'src')
+  it('SERVER_DATA_DIR point to the right place', async function () {
+    const testPaths = [
+      'treasury',
+      env.EMPTY_TEMPLATE_NAME
+    ]
+
+    for (const relative of testPaths) {
+      const absolute = path.join(env.SERVER_DATA_DIR, relative)
+      const pathExists = await exists(absolute)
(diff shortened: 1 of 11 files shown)


tests/server/environment.spec.js
```diff
@@ -1,14 +1,40 @@
...
+
+    for (const relative of testPaths) {
+      const absolute = path.join(env.SERVER_CODE_DIR, relative)
+      const pathExists = await exists(absolute)
+      assert.ok(pathExists, `expected ${absolute} to exist, is SERVER_CODE_DIR wrong?`)
+    }
   })
 
-  it('points the CODE_DIR at src', function () {
-    const env = require(underTest)
```
ajhyndman, line 11: Oh, I'd guess this was to support updating env values while running tests in watch mode without restarting the script.

I've never tried to do that, so this seems fine to me.

tests/server/mocha_init.js
```diff
@@ -8,8 +8,16 @@ const { setupAgencies } = require('./fixtures/add-dummy-data')
...
+global.requireSrc = function requireSrc (fpath) {
+  return require(
+    fpath
+      .replace(/(\.[^.]*)*\.spec/, '')
+      // GOST repo path of ARPA Reporter code
+      .replace(/\/__tests__\/arpa_reporter\/server\//, '/src/arpa_reporter/')
+      // Legacy ARPA repo paths
+      .replace(/\/tests\//, '/src/')
+  )
+}
```
ajhyndman, line 20: Ohhh, this is where that magic import rewriting is set up!  I thought it was a weird feature for mocha to ship built-in.

ajhyndman (commented):
LGTM

Note: It should be fine to re-number the seeds if you run into any issues.  Seeds only run when triggered manually, and we shouldn't need to re-run seeds on any existing deployments.

ajhyndman (approved):

PR #516, opened 2022-09-07
GOST integration: script to dump all relevant database tables and all uploaded files to a zip

mattbroussard:
This is one part of what will be used to migrate existing partners with their own Render instances to a unified ARPA-Reporter-in-GOST world. The idea is that you'd:
1. run this script on their ARPA instance
2. temporarily enable SSH on both the ARPA instance and GOST prod instance
3. scp/rsync the file from ARPA to GOST
4. disable SSH
5. run a corresponding import script on the GOST side (implemented in https://github.com/usdigitalresponse/usdr-gost/pull/312)
6. delete the zip files on both sides

One concern about this script is that `adm-zip` reads the contents of all uploaded files into memory before writing the output zip. I'm not sure how many/how big uploads are practically dealing with, though I'm guessing this probably won't be a big issue?

There is also one minor drive-by fix in this PR: adding `ensureAsyncContext` around the template upload endpoint's `multer` middleware -- this was needed in order to test that the script correctly picked up uploaded templates (and indeed without this fix, uploading custom templates is currently broken since #473)

### DIFF
--- a/scripts/dump_data.js
+++ b/scripts/dump_data.js
@@ -1,0 +1,119 @@
+const _ = require("lodash");
+const AdmZip = require("adm-zip");
+const fs = require("fs/promises");
+const inquirer = require("inquirer");
+const path = require("path");
+
+const knex = require("../src/server/db/connection");
+const { uploadFSName } = require("../src/server/services/persist-upload");
+const { periodTemplatePath } = require("../src/server/services/get-template");
+
+const TABLES = [
+  "users",
+  "agencies",
+  "reporting_periods",
+  "uploads", // has FK to reporting_periods
+  "arpa_subrecipients", // has FK to uploads
+  "application_settings", // has FK to reporting_periods
+  "projects", // has FK to reporting_periods
+  "period_summaries", // has FK to reporting_periods, projects
+];
+
+async function getDatabaseContents() {
+  console.log("Fetching all DB rows...");
+  const start = Date.now();
+
+  const data = await Promise.all(
+    TABLES.map((tableName) => knex(tableName).select("*"))
+  );
+  const ret = _.fromPairs(_.zip(TABLES, data));
+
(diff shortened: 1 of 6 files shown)


scripts/dump_data.js
```diff
@@ -0,0 +1,167 @@
...
+];
+
+async function getDatabaseContents() {
+  console.log("Fetching all DB rows...");
+  const start = Date.now();
+
+  const data = await Promise.all(
+    TABLES.map((tableName) => knex(tableName).select("*"))
+  );
+  const ret = _.fromPairs(_.zip(TABLES, data));
```
trumanc, line 31: It might be a useful/simple sanity check to print out the number of rows in each table here, just so we can quickly compare it to a similar printout at the import side

trumanc (approved):
This side seems simple enough! Heading over to the import PR now...

scripts/dump_data.js
```diff
@@ -0,0 +1,167 @@
...
+  "uploads", // has FK to reporting_periods
+  "arpa_subrecipients", // has FK to uploads
+  "application_settings", // has FK to reporting_periods
+  "projects", // has FK to reporting_periods
+  "period_summaries", // has FK to reporting_periods, projects
+];
+
+async function getDatabaseContents() {
+  console.log("Fetching all DB rows...");
+  const start = Date.now();
```
ajhyndman, line 26: Nit: The [performance.now()](https://nodejs.org/api/perf_hooks.html#performancenow) API is available in Node now, and gives slightly higher resolution timestamps with slightly less overhead.

ajhyndman (approved):
Yeah, this looks great.  👍

scripts/dump_data.js
mattbroussard: Oh, good to know.

PR #518, opened 2022-09-08
GOST Integration: populate redirect_to field during login

mattbroussard:
We set a new URL param, `redirect_to` when redirecting to the login page from another VueRouter route, and the login page forwards that to its `/api/sessions` call.

In standalone ARPA Reporter repo, this parameter is ignored. But in GOST once https://github.com/usdigitalresponse/usdr-gost/pull/288 is merged, it will cause the login link email that's sent to also include a `redirect_to` field. In this way, after logging in using the GOST `/api/sessions` endpoints, the user is redirected back to ARPA Reporter rather than going to the GOST/ID Tool dashboard.

The VueRouter APIs used here automatically account for the `base` config option that will be set in GOST to nest the ARPA Reporter app under a URL path.

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -124,8 +124,12 @@ function loggedIn () {
 router.beforeEach((to, from, next) => {
   if (to.matched.some(record => record.meta.requiresLogin)) {
     if (!loggedIn()) {
+      // This will include any router base URL, if configured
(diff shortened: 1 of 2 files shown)


ajhyndman (approved):
Looks great

PR #520, opened 2022-09-12
Bugbash fix: export useUser

mattbroussard:
This is referenced in a DB shim (see [here](https://gist.github.com/mattbroussard/7f725a9090ee5a9ff9be77eaab555e76)), but was not actually exported.


ajhyndman (approved):
Whoops, I'm sure I meant to export that.

PR #521, opened 2022-09-13
Add warning/error messages on login screen

trumanc:
This PR adds a couple simple error/warning alerts that are configurable via env variable.
It supports a warning message that still allows the user to login (in case they want to sign in and access their old data/reports, as well as an error version that blocks the login process entirely.

The messages themself are set in the env variable, so we can include useful info like the url for the new tool, and reminders to update their bookmarks, etc...

If login is fully disabled:
Screen Shot 2022-09-13 at 12 23 51 PM

If it's only a warning message:
Screen Shot 2022-09-13 at 12 22 50 PM


src/server/routes/sessions.js
```diff
@@ -47,6 +48,12 @@ router.get('/logout', function (req, res) {
 })
 
 router.post('/', async function (req, res, next) {
+  if (LOGIN_DISABLED_MESSAGE) {
+    res.json({
+      success: false,
+      message: `Login disabled: ${LOGIN_DISABLED_MESSAGE}`
+    })
```
ajhyndman, line 55:
```suggestion
    return res.json({
      success: false,
      message: `Login disabled: ${LOGIN_DISABLED_MESSAGE}`
    })
```

I assume you mean to actually prevent sending the email in this case?  In which case, this line should return.

ajhyndman (changes requested):

ajhyndman:
Thanks for putting this together!

mattbroussard:
One thought is it would be nice for the URL to be a clickable link. Could either use something like [this](https://www.npmjs.com/package/linkifyjs) to linkify via regex, make the env variable be HTML, or change the structure of the message so the text and URL are in separate variables.

trumanc:
@mattbroussard  I considered making the link clickable, but decided against it for a couple minor reasons:
First: I wasn't thrilled with any if the implementation options (installing a new dependency for a temporary feature seems like overkill, raw html seems error prone, as does making the message and url separate env variables that could be sent separately.

Second: I'm actually not mad about the user having to copy/paste the link, because if they're seeing this message it probably means that they followed an out-of-date bookmark that they should edit anyways. The more painful this 'follow-the-link' process is, the more likely they are to update the bookmark and not make the mistake again in the future.

If you feel strongly about making it clickable I can definitely add it though! Just let me know

ajhyndman (approved):

PR #522, opened 2022-09-13
Display reporting period name, not id

trumanc:
Updates the upload screen to use the reporting period name instead of the id in the. two places it is dipslayed. This will become a bigger concern as we lean in to multi-tenancy in the GOST repo and period ids will be significantly higher than they currently are.

Old screenshot (says "period 1" in the rows of the left column, and the title of the right column)
Screen Shot 2022-09-13 at 1 22 33 PM

Replaced both instances in this updated screenshot
Screen Shot 2022-09-13 at 1 21 34 PM

### DIFF
--- a/src/views/Upload.vue
+++ b/src/views/Upload.vue
@@ -52,7 +52,7 @@
 
           <li class="list-group-item">
             <span class="font-weight-bold">Reporting Period: </span>
-            {{ upload.reporting_period_id }}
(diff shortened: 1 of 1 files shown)


mattbroussard (approved):

mattbroussard:
Gonna go ahead and merge this so there are aren't open PRs when we do repo merge tomorrow.

